PR #22, opened 2022-07-14
'Solution'

daria-niklaevna:


katevasilenko (changes requested):
it seems like you forgot to add migrations to your PR

cinema/views.py
```diff
@@ -1 +1,51 @@
...
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all().select_related("movie", "cinema_hall")
```
MasterpieceElbow, line 40: You can use `select_related` and `prefetch_related` without `.all()`.

MasterpieceElbow (approved):

PR #23, opened 2022-07-14
Katya's Solution

katevasilenko:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return self.first_name + " " + self.last_name
```
DmytroSvirsa, line 32: i think f string better

cinema/serializers.py
```diff
@@ -1 +1,87 @@
...
+        )
+
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.StringRelatedField(
+        read_only=True,
+        many=True
+    )
+    actors = serializers.StringRelatedField(
+        many=True
```
DmytroSvirsa, line 47: read_only=True ?

cinema/serializers.py
```diff
@@ -1 +1,87 @@
...
+            "title",
+            "description",
+            "duration",
+            "genres",
+            "actors"
+        )
+
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.StringRelatedField(
```
DmytroSvirsa, line 42: SlugRelatedField looks more relieble

DmytroSvirsa (approved):

cinema/models.py
katevasilenko: Yeah, I prefer it too, but just made in the same style as other code

TPolina (approved):

PR #33, opened 2022-07-14
Solution

BohdanKryven:


KatiukhaO (changes requested):
Tests do not pass on Git Hub, because you need to add debug tool bar in requirement txt and then run this file on Git Hub.
Or remove debug tool bar from settings

BohdanKryven:
> Tests do not pass on Git Hub, because you need to add debug tool bar in requirement txt and then run this file on Git Hub. Or remove debug tool bar from settings

Thanks! Fixed

KatiukhaO (approved):

PR #73, opened 2022-11-17
Solution

olia-trofymchuk:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
mltdll, line 35: These two methods do the same thing. Maybe it's better to use one of them in the other?

mltdll (changes requested):

cinema/models.py
olia-trofymchuk: Thank u! I'll try.

mltdll (approved):

cinema/models.py
```diff
@@ -27,8 +27,12 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
```
Nattalli, line 34: what about type annotations?

Nattalli (changes requested):

cinema/models.py
olia-trofymchuk: okaayyyyy :) i forgot)))

Nattalli: unfortunately, almost like everyone else. That's why it's ok 😸

PR #75, opened 2022-11-17
Solution

Yaroslav-Stetsko:


HStakhiv (approved):
Looks good. But when you use `fields = "__all__"`, the output will be different from in the example.

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```

Nattalli (changes requested):

cinema/models.py
Yaroslav-Stetsko: done

Nattalli (approved):

PR #76, opened 2022-11-17
Solution

HStakhiv:


cinema/views.py
```diff
@@ -1 +1,71 @@
...
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+
+    def get_queryset(self):
+        queryset = self.queryset
+
+        if self.action in ("list", "retrieve"):
+            queryset = queryset.prefetch_related("movie", "cinema_hall")
```
VadymShkarbul, line 60: Maybe you should use select_related here (but django-debug-toolbar says that no matter)

VadymShkarbul (approved):

cinema/views.py
HStakhiv: Thanks. Yes

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```

Nattalli (changes requested):

cinema/models.py
HStakhiv: Sorry. I do not agree with point number 2. The two methods are the same, but if I leave, for example, only the str method and change it in the future, then I will lose the functionality of the full_name field.  I want full_name to be independent, so I created this property.

HStakhiv: Sorry. I realized. I'll just do `return full_name()` in `str`
Thanks

Nattalli: btw, I partially agree about independence. But if you leave these methods independent, then they should at least be written in the same way inside (adhere to the same style)

Nattalli (approved):

PR #83, opened 2022-11-18
Solution

Kitazuka:


IvanStored (changes requested):
Solve `N+1 problem` using `select\prefetch related`

IvanStored (approved):
gj!

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
Nattalli, line 35: These two methods do the same thing. Maybe it's better to use one of them in the other? Or, at least, you can make them in one style

Nattalli (changes requested):

TPolina (approved):

PR #84, opened 2022-11-18
Solution

linaSam:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
wQuelS, line 35: Seems like str method here is doing exact same thing as the full_name method. Dunno if it is a mistake, but I'd have changed the logic a bit in my solution.

cinema/views.py
```diff
@@ -1 +1,51 @@
-# write views here
+from rest_framework import viewsets
+
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
+from cinema.serializers import GenreSerializer, ActorSerializer, \
+    CinemaHallSerializer, MovieSerializer, MovieSessionSerializer, \
```
wQuelS, line 5: Import style looks not so good. 
Try black module. [https://pypi.org/project/black/](url)

cinema/views.py
```diff
@@ -1 +1,51 @@
...
+    serializer_class = ActorSerializer
+
+
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all().prefetch_related("genres", "actors")
```
wQuelS, line 26: you can remove all() here

cinema/views.py
```diff
@@ -1 +1,51 @@
...
+        if self.action == "list":
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all().select_related(
```
wQuelS, line 39: same

wQuelS (changes requested):
Left some comments. Also, you have additional security issues due to db itself in your PR. Please, remove it.

cinema/models.py
linaSam: return full_name in __str__? I agree that it reduces unnecessary code, IDK whether it's a mistake or not. I guess we will know soon:)

wQuelS (approved):
Looks nice

cinema/models.py
```diff
@@ -27,8 +27,12 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
-        return self.first_name + " " + self.last_name
+        return self.full_name
```
2) that was a good point about using one method in another one

Nattalli (changes requested):

Nattalli (approved):

PR #85, opened 2022-11-18
Solution

VadymShkarbul:


HStakhiv (approved):
Looks good

cinema/models.py
```diff
@@ -12,24 +12,28 @@ class CinemaHall(models.Model):
...
     last_name = models.CharField(max_length=255)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return self.first_name + " " + self.last_name
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
```
Nattalli, line 36: These two methods do the same thing. Maybe it's better to use one of them in the other?

cinema/serializers.py
```diff
@@ -1 +1,88 @@
...
+        )
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
+    movie = MovieListSerializer(read_only=True)
+    cinema_hall = CinemaHallSerializer(read_only=True)
+
+    class Meta:
+        model = MovieSession
+        fields = ("id", "show_time", "movie", "cinema_hall")
```
Nattalli, line 88: you have the same Meta class as in the `MovieSessionSerializer`. This class is inherited from the `MovieSessionSerializer`. So I think that you can remove the Meta class in this serializer

cinema/urls.py
```diff
@@ -1 +1,21 @@
...
+
+router = routers.DefaultRouter()
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("cinema/", include(router.urls)),
```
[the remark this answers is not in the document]
Nattalli: hm, I agree. That is not an error, but it's better to add `api/cinema/`in the root folder

Nattalli (changes requested):

cinema/models.py
VadymShkarbul: In this case yes, but I thought about separate use of this both methods. If in future we want to add some new field to the actors model and override str representation (for example in admin panel etc. ` @property
    def full_name(self) -> str:
        return f"{self.first_name} {self.last_name}"

    def __str__(self) -> str:
        return self.full_name`

Nattalli: hmm, agree with you. if you leave these methods independent, then they should at least be written in the same way inside (adhere to the same style)

Nattalli (approved):
gj, and you was right about those methods

PR #88, opened 2022-11-19
Solution

VadimSabadyr:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
wQuelS, line 35: Seems like str method here is doing exact same thing as the full_name method. Dunno if it is a mistake, but I'd have changed the logic a bit in my solution.

cinema/views.py
```diff
@@ -1 +1,61 @@
...
+    serializer_class = ActorSerializer
+
+
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all().prefetch_related("genres", "actors")
```
wQuelS, line 32: You don't need all() here

wQuelS (changes requested):

cinema/models.py
```diff
@@ -27,8 +27,9 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
-    def __str__(self):
-        return self.first_name + " " + self.last_name
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
```
Nattalli, line 32: 1) add a type annotation
2) why did you you remove the `__str__` method?

cinema/serializers.py
```diff
@@ -1 +1,89 @@
...
+        )
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
+    movie = MovieListSerializer(read_only=True)
+    cinema_hall = CinemaHallSerializer(read_only=True)
+
+    class Meta:
+        model = MovieSession
+        fields = ("id", "show_time", "movie", "cinema_hall")
```
Nattalli, line 89: you have the same Meta class as in the `MovieSessionSerializer`. This class is inherited from the `MovieSessionSerializer`. So I think that you can remove the Meta class in this serializer

Nattalli (changes requested):

Nattalli (approved):
👍🏻

PR #89, opened 2022-11-19
Solution

Vasyl-Poremchuk:


cinema/serializers.py
```diff
@@ -1 +1,106 @@
...
+        slug_field="name",
+    )
+    actors = serializers.SlugRelatedField(
+        many=True,
+        read_only=True,
+        slug_field="full_name",
+    )
+
+    class Meta:
+        model = Movie
```
wQuelS, line 62: Your MovieListSerializer inherits from MovieSerializer. Do you really need class Meta here with exact same fields ?

cinema/views.py
```diff
@@ -1 +1,81 @@
...
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+
+        if self.action == "list":
```
wQuelS, line 45: Please check for similar or duplicate queries with debug_toolbar when you retrieving detail info about movie

cinema/views.py
```diff
@@ -1 +1,81 @@
...
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+
+        if self.action == "list":
```
wQuelS, line 68: same, what about N+1 problem when retrieving movie_sessions detail info?

wQuelS (changes requested):

wQuelS (approved):
Well done!

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```

cinema/serializers.py
```diff
@@ -1 +1,98 @@
...
+    MovieSession,
+)
+
+
+class ActorSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Actor
+        fields = (
+            "id", "first_name",
+            "last_name", "full_name",
```
Nattalli, line 17: hmm, did you run `black`? It doesn't look good to me, but I can't tell that it's a mistake

Nattalli: just maybe 
1) 
```python
fields = ("a", "b", "c", "d")
```
or 
2) 
```python
fields = (
"a",
"b",
"c",
"d"
)
```
would that be better? WDYT?

Nattalli (changes requested):

cinema/models.py
Vasyl-Poremchuk: Oh, I didn't notice it :(

cinema/serializers.py
Vasyl-Poremchuk: I agree with you, the second option will be better in this case)

Nattalli (approved):

PR #91, opened 2022-11-19
Develop

alexzubkoff:


cinema/tests/test_movie_api.py
```diff
@@ -70,8 +70,8 @@ def test_post_invalid_movies(self):
             },
         )
         superman_movies = Movie.objects.filter(title="Superman")
-        self.assertEqual(movies.status_code, status.HTTP_400_BAD_REQUEST)
-        self.assertEqual(superman_movies.count(), 0)
+        self.assertEqual(movies.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(superman_movies.count(), 1)
```
mltdll, line 74: You are not supposed to change the tests. Instead, the `actors` and `genres` fields should be required.

cinema/urls.py
```diff
@@ -1 +1,117 @@
...
+)
+
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
```
mltdll, line 19: There is a way to add a single `path` to the `urlpatterns` using the `router` created above.
Right now it's not even used at all.

cinema/views.py
```diff
@@ -1 +1,63 @@
...
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+
+        if self.action != "list":
```
mltdll, line 37: Doesn't the list display need query optimization?

cinema/views.py
```diff
@@ -1 +1,63 @@
...
+            queryset = queryset.prefetch_related("genres").\
+                prefetch_related("actors")
```
mltdll, line 39: It's possible to pass two arguments to `.prefetch_related()` instead.

By the way, splitting lines with `()` is the preferred style.

cinema/views.py
```diff
@@ -1 +1,63 @@
...
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+
+        if self.action != "list":
+            queryset = queryset.select_related("movie").\
+                select_related("cinema_hall")
+        return queryset
```
mltdll, line 58: Same comments as the two above.

mltdll (changes requested):

alexzubkoff:
Thank for a good advices!

cinema/tests/test_movie_api.py
mltdll: This comment was among the most important about the whole review. I'd suggest fixing it after all. :)

mltdll (changes requested):

PR #94, opened 2022-11-20
Solution

matechris:


VadymShkarbul (approved):
I can't fault anything except the type annotations.

cinema/models.py
```diff
@@ -12,22 +12,26 @@ class CinemaHall(models.Model):
...
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
-    def __str__(self):
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
+    def __str__(self) -> str:
         return self.first_name + " " + self.last_name
```
Nattalli, line 35: These two methods do the same thing. Maybe it's better to use one of them in the other?

Nattalli (changes requested):

Nattalli (approved):

PR #95, opened 2022-11-20
Solution!

ketstap162:


cinema/views.py
```diff
@@ -1 +1,67 @@
...
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.select_related("cinema_hall", "movie")
+    serializer_class = MovieSessionSerializer
+
+    def get_queryset(self):
+        if self.action == "list":
+            return self.queryset
+        if self.action == "retrieve":
+            return self.queryset.prefetch_related(
```
Vasyl-Poremchuk, line 54: In this case it doesn't work, you also have 7 queries in your DB, you should use `select_related()` here and you will have 5 queries in your DB

Vasyl-Poremchuk (changes requested):

cinema/views.py
ketstap162: I have static numb of queries with prefetch_related. I use info about many actors and many genres. So, select_related is bad for this situation. It's my opinion...
image

Vasyl-Poremchuk: Ok, I got you. I think the mentor will give us the answer, which is better to use in this case)

Vasyl-Poremchuk (approved):

cinema/models.py
```diff
@@ -27,9 +27,13 @@ class Actor(models.Model):
...
     last_name = models.CharField(max_length=255)
 
-    def __str__(self):
+    @property
+    def full_name(self):
         return self.first_name + " " + self.last_name
 
+    def __str__(self):
+        return self.full_name
+
```
Nattalli, line 36: add a type annotation

cinema/models.py
```diff
@@ -38,6 +42,13 @@ class Movie(models.Model):
     genres = models.ManyToManyField(Genre)
     actors = models.ManyToManyField(Actor)
```
Nattalli, line 43: add related_name

cinema/views.py
Nattalli: @ketstap162 I did not quite understand where you got your statistics. Tell me in more detail why you decided that way?

Nattalli (changes requested):

cinema/views.py
ketstap162: There are no statistics, I just learned that prefetch_related is used for ManyToMany relation. I discussed this with colleagues, the django interface provides this under the hood (if I'm not mistaken), but I would like to hear the exact version from you

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.select_related("cinema_hall", "movie")
```
Y-Havryliv, line 47: you don't need `get_queryset` now

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Nice!

PR #99, opened 2022-11-20
Solution

Ilyakson:


cinema/views.py
```diff
@@ -1 +1,58 @@
...
+    serializer_class = GenreSerializer
+
+
+class ActorViewSet(viewsets.ModelViewSet):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all().prefetch_related("genres", "actors")
```
mltdll, line 33: Using `.all()` is not necessary here.

cinema/views.py
```diff
@@ -1 +1,58 @@
...
+            return MovieDetailSerializer
+
+        if self.action == "list":
+            return MovieListSerializer
+
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all().select_related(
```
mltdll, line 46: Same as above.

mltdll (changes requested):

mltdll (approved):
Actually, this is really a minor thing, and, taking into account the blackouts, I think it's better to spend time on something more important than removing `.all()` from a pull request.
Therefore, GJ.
I think this solution should be approved.

cinema/views.py
Nattalli: yes, remove it

cinema/views.py
Nattalli: and here

Nattalli (changes requested):

TPolina (approved):

PR #102, opened 2022-11-21
Solution

misha-doroshenko:


cinema/views.py
```diff
@@ -1 +1,56 @@
...
+            return MovieListSerializer
+
+        if self.action in "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
```
KatrinLazarenko, line 47: Maybe here better to redefine queryset using prefetch/select_related for list and retrieve for reducing same queries?  The same in MovieViewSet

KatrinLazarenko (changes requested):

KatrinLazarenko (approved):
Looks nice

TPolina (approved):

PR #106, opened 2022-11-21
Solution

rostyak:


Aleksandro777 (approved):
Nice

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
Nattalli, line 35: add a type annotation 🥲

Nattalli: These two methods do the same thing. Maybe it's better to use one of them in the other?
Or at least you can make it in one style

cinema/serializers.py
```diff
@@ -1 +1,102 @@
...
+    cinema_hall = CinemaHallSerializer(many=False)
+
+    class Meta:
+        model = MovieSession
+        fields = (
+            "id",
+            "show_time",
+            "movie",
+            "cinema_hall",
+        )
```
Nattalli, line 102: you have the same Meta class as in the `MovieSessionSerializer`. This class is inherited from the `MovieSessionSerializer`. So I think that you can remove the Meta class in this serializer

Nattalli (changes requested):

Nattalli (approved):
👍🏻

PR #124, opened 2023-01-10
solution

hikehikehike:


cinema/views.py
```diff
@@ -1 +1,76 @@
...
+    serializer_class = GenreSerializer
+
+
+class ActorViewSet(viewsets.ModelViewSet):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all().prefetch_related("genres", "actors")
```
YuriiKindrat, line 43: image
fix this

YuriiKindrat (changes requested):
remove db from ur PR

cinema/views.py
hikehikehike: done

andrewbelo (approved):

PR #133, opened 2023-02-07
Solution

KatyaVasylieva:


cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi, line 22: No need to include when you do not use basepath just:
urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

stanhjr (approved):

PR #134, opened 2023-02-09
API Serializers

anton-chumak-main:
Implement serializers and views

### DIFF
--- a/cinema/models.py
+++ b/cinema/models.py
@@ -30,6 +30,10 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
 
 class Movie(models.Model):
     title = models.CharField(max_length=255)
--- a/cinema/serializers.py
+++ b/cinema/serializers.py
@@ -1,1 +1,81 @@
-# write serializers here
+from rest_framework import serializers
+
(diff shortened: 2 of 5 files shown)


cinema/models.py
```diff
@@ -30,6 +30,10 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
```
Bohdan-Salamakha, line 35: why different code style in same logic?
```
    def __str__(self):

    @property
    def full_name(self) -> str:
        return f"{self.first_name} {self.last_name}"
```

Bohdan-Salamakha (changes requested):
v1

cinema/models.py
```diff
@@ -28,7 +28,11 @@ class Actor(models.Model):
     last_name = models.CharField(max_length=255)
 
     def __str__(self):
-        return self.first_name + " " + self.last_name
+        return f"{self.first_name} {self.last_name}"
```
olena-hul, line 31: Use self.full_name

olena-hul (changes requested):

cinema/models.py
anton-chumak-main: but I don't have such a field ?

olena-hul (approved):

PR #147, opened 2023-03-14
Solution

darynka-h:


cinema/urls.py
```diff
@@ -1 +1,25 @@
...
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+
+]
```
ArsenPidhoretskyi, line 23: No need to include here, because base path not provided. SImplify this:
urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

PR #151, opened 2023-03-19
Solution

IvanMayor:


cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi, line 21: No need to include if basepath is not gived
Use: urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #152, opened 2023-03-20
Api Serializers

MarynaProkhorenko:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return self.first_name + " " + self.last_name
```
Grayder0152, line 32: use f-string in this case

Grayder0152 (approved):

Grayder0152 (changes requested):
Pls fix tests

cinema/models.py
vasylhnatiuk: not a @property, use f"{self.firts_name} {self.last_name}"

vasylhnatiuk (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #153, opened 2023-03-21
Solution

MykytaMoshchenko:


vasylhnatiuk (changes requested):

cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = routers.DefaultRouter()
+router.register("actors", ActorViewSet)
+router.register("genres", GenreViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi, line 21: No need to use include in this case when we do not have basepath
Simplify it by: urlpatterns = router.urls

cinema/views.py
```diff
@@ -1 +1,69 @@
...
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.prefetch_related("genres", "actors")
+    serializer_class = MovieSerializer
+
+    def get_serializer_class(self):
```
ArsenPidhoretskyi, line 45: Provide here also type annotations

ArsenPidhoretskyi (changes requested):

katevasilenko (changes requested):
Looks good to me but could you remove db from your PR?

katevasilenko (approved):
Good job, thanks for quick update

PR #160, opened 2023-04-02
Solution

MaybeTI:


AndereLion (approved):

cinema/serializers.py
```diff
@@ -35,18 +35,16 @@ class MovieListSerializer(MovieSerializer):
     )
     actors = serializers.StringRelatedField(many=True, read_only=True)
 
-    class Meta:
+    class Meta(MovieSerializer.Meta):
```
wQuelS, line 38: you don't need to inherit here in Meta. Your serializer inherits from `MovieSerializer` and it is enough for it to have all fields from parent class

cinema/serializers.py
```diff
@@ -35,18 +35,16 @@ class MovieListSerializer(MovieSerializer):
...
         model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
 
 
 class MovieDetailSerializer(MovieSerializer):
     genres = GenreSerializer(many=True, read_only=True)
     actors = ActorSerializer(many=True, read_only=True)
 
-    class Meta:
+    class Meta(MovieSerializer.Meta):
```
wQuelS, line 46: same

cinema/views.py
```diff
@@ -43,7 +43,6 @@ def get_serializer_class(self):
 
 class MovieSessionViewSet(viewsets.ModelViewSet):
     queryset = MovieSession.objects.all()
-    serializer_class = MovieSessionSerializer
```
wQuelS: ok, you've decided to remove this attr.. Another solution was on line 54 write `return self.serializer_class`

wQuelS (changes requested):

cinema/views.py
```diff
@@ -1 +1,55 @@
...
+            return MovieListSerializer
+
+        elif self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
```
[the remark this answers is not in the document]
YuriiKindrat: u didn't fix that, u are using related tables in ur serializers

YuriiKindrat (changes requested):

PR #174, opened 2023-04-06
Solution

inrock1:


cinema/models.py
```diff
@@ -30,6 +30,10 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
```
WDemidenko, line 32: you could use self.full_name here

WDemidenko (approved):
Well done!

cinema/models.py
inrock1: really. thanks

PR #187, opened 2023-04-06
Solution

ostapT:


me1nyk (approved):

cinema/models.py
```diff
@@ -12,22 +12,26 @@ class CinemaHall(models.Model):
...
 
 
 class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
-    def __str__(self):
+    @property
+    def full_name(self) -> str:
+        return self.first_name + " " + self.last_name
```
wQuelS, line 32: use f-string. ?fix everywhere

cinema/views.py
```diff
@@ -1 +1,55 @@
...
+    serializer_class = ActorSerializer
+
+
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.prefetch_related("actors", "genres")
```
wQuelS, line 33: it is better to use `prefetch_related` if specific action occurs, since you don't have N+1 problem in other cases.

cinema/views.py
```diff
@@ -1 +1,55 @@
...
+
+    def get_serializer_class(self):
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        if self.action == "list":
+            return MovieListSerializer
+
+        return MovieSerializer
```
wQuelS, line 42: `return self.serializer_class` is better according to best practices. Fix everywhere

cinema/views.py
```diff
@@ -1 +1,55 @@
...
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.select_related("movie", "cinema_hall")
```
wQuelS, line 46: again, use it on specific action

cinema/views.py
```diff
@@ -1 +1,55 @@
...
+
+    def get_serializer_class(self):
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+
+        if self.action == "list":
+            return MovieSessionListSerializer
+
+        return MovieSessionSerializer
```
wQuelS, line 55: same with `self.serializer_class`

wQuelS (changes requested):

wQuelS (approved):

PR #194, opened 2023-04-06
 Solution

tytomyr:


viktoria-rybenchuk (approved):
GJ

cinema/models.py
```diff
@@ -30,6 +30,10 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    @property
+    def full_name(self):
```
wQuelS, line 34: type annotations?

cinema/models.py
```diff
@@ -30,6 +30,10 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
```
wQuelS, line 31: f-string is better than string concatenation

cinema/views.py
```diff
@@ -1 +1,81 @@
...
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
```
wQuelS, line 43: `return self.serializer_class` is better

cinema/views.py
```diff
@@ -1 +1,81 @@
...
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+
+        return MovieSessionSerializer
```
wQuelS, line 71: same

wQuelS (changes requested):

cinema/views.py
tytomyr: in this case, serializer_class is a MovieDetailSerializer, so returning this serializer will cause an error

wQuelS (approved):

PR #198, opened 2023-04-06
Solution

me1nyk:


cinema/views.py
```diff
@@ -1 +1,65 @@
...
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
```
YuriiKindrat, line 37: no need to create variable for attribute

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+            return MovieDetailSerializer
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
```
YuriiKindrat, line 55: same

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+        if self.action in ("list", "retrieve"):
+            queryset = queryset.prefetch_related("actors", "genres")
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
```
YuriiKindrat, line 47: u can use self.serializer_class here

YuriiKindrat (changes requested):

PR #200, opened 2023-04-06
Solution

ada-krav:


cinema/serializers.py
```diff
@@ -1 +1,94 @@
...
+            "description",
+            "duration",
+            "genres",
+            "actors"
+        )
+
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.SerializerMethodField()
+    actors = serializers.SerializerMethodField()
```
YuriiKindrat, line 44: u can use StringRelated or SlugRelated fields here, to simplify ur code

cinema/views.py
```diff
@@ -1 +1,51 @@
...
+    serializer_class = ActorSerializer
+
+
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.prefetch_related("genres", "actors")
```
YuriiKindrat, line 33: do u need an optimized query in ur actions except list, retrieve ?

cinema/views.py
```diff
@@ -1 +1,51 @@
...
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.prefetch_related("cinema_hall")
```
YuriiKindrat, line 44: does it has m2m relations with cinema_hall? also what about movie ?
also the same for optimized query as above

YuriiKindrat (changes requested):

wQuelS (changes requested):

PR #204, opened 2023-04-07
Solution

ZhAlexR:


cinema/serializers.py
```diff
@@ -1 +1,101 @@
...
+class MovieSessionListSerializer(MovieSessionSerializer):
+    movie_title = serializers.CharField(source="movie.title", max_length=255)
+    cinema_hall_name = serializers.CharField(
+        source="cinema_hall.name", max_length=255
+    )
+    cinema_hall_capacity = serializers.IntegerField(
+        source="cinema_hall.capacity"
+    )
+
+    class Meta(MovieSessionSerializer.Meta):
```
YuriiKindrat, line 75: why do u need to inherit from MovieSessionSerializer and Meta here ?

cinema/serializers.py
```diff
@@ -1 +1,101 @@
...
+            "cinema_hall_capacity",
+        )
+        read_only_fields = [
+            "movie_title",
+            "cinema_hall_name",
+            "cinema_hall_capacity",
+        ]
+
+
+class MovieSessionDetailSerializer(serializers.ModelSerializer):
```
YuriiKindrat, line 90: why don't u inherit from MovieSessionSerializer to avoid implementing meta class here?

cinema/views.py
```diff
@@ -1 +1,62 @@
...
+    def get_serializer_class(self) -> Type[Serializer]:
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.select_related("movie", "cinema_hall")
```
YuriiKindrat, line 55: do u need to optimize query in actions except list, retrieve ?

cinema/serializers.py
ZhAlexR: > why do u need to inherit from MovieSessionSerializer and Meta here ?

cinema/serializers.py
ZhAlexR: > why don't u inherit from MovieSessionSerializer to avoid implementing meta class here?

Agree. It is redundant

YuriiKindrat (approved):

PR #208, opened 2023-04-08
Solution

Yevheniy-Lototskiy:


cinema/views.py
```diff
@@ -1 +1,65 @@
...
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
```
YuriiKindrat, line 37: no need to create variable for attribute )

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
```
YuriiKindrat, line 51: u can use here, self.serializer_class

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+    serializer_class = MovieSessionSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+
+        return MovieSessionSerializer
```
YuriiKindrat, line 65: same here

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
```
YuriiKindrat, line 55: what about optimizing query here?

YuriiKindrat (changes requested):

PR #209, opened 2023-04-08
Solution

Zenionoi:


Zenionoi:
In second commit fixed flake8 error with black

cinema/serializers.py
```diff
@@ -1 +1,81 @@
...
+
+class ActorSerializer(serializers.ModelSerializer):
+    full_name = serializers.CharField(source="__str__", read_only=True)
+
+    class Meta:
+        model = Actor
+        fields = ("id", "first_name", "last_name", "full_name")
+
+    def get_full_name(self, obj):
+        return str(obj)
```
YuriiKindrat, line 14: why do u need this ?

YuriiKindrat (changes requested):

cinema/serializers.py
Zenionoi: I saw in one tutorial about the definition of the get_<field_name> method, and you are right, it is not used here

PR #211, opened 2023-04-09
Develop

ghost:


cinema/views.py
```diff
@@ -1 +1,70 @@
...
+
+
+class MovieSessionViewSet(ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+        if self.action in ("list", "retrieve"):
+            queryset = queryset.prefetch_related("movie", "cinema_hall")
```
YuriiKindrat, line 62: does it has m2m relations to movie and cinema_hall ?

YuriiKindrat (changes requested):

cinema/views.py
oksana0410: No, just foreign key

YuriiKindrat (approved):

PR #215, opened 2023-04-26
Solution

Oleksiy-Liubchenko:


cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
ArsenPidhoretskyi, line 21: No need to use include if basepath is not provided use:
urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #226, opened 2023-05-26
Solution

Amirammi:


cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router.register("genres", GenreViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("movies", MovieViewSet)
+router.register("actors", ActorViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
+
```
ArsenPidhoretskyi, line 22: No need to include if base path is not provided you can do:
urlpatters = router.urls

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done

PR #229, opened 2023-05-29
solution

megah1t:


cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = routers.DefaultRouter()
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi, line 21: No need to use include if basepath is not exist. 
Instead do: urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done

PR #234, opened 2023-06-06
Solution

anastasiiashchoholieva:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return self.first_name + " " + self.last_name
```
EdAlekseiev, line 32: Use f-string instead of string concatenation. Add type hint for returning value

cinema/views.py
```diff
@@ -1 +1,57 @@
...
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.prefetch_related("actors", "genres")
+    serializer_class = MovieSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
```
EdAlekseiev, line 26:
```suggestion
        return super().get_serializer_class()
```

cinema/views.py
```diff
@@ -1 +1,57 @@
...
+    )
+    serializer_class = MovieSessionSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+        return MovieSessionSerializer
```
EdAlekseiev, line 57: The same issue as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #244, opened 2023-06-25
solution

IvanKorshunovE:


cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+router.register("movies", MovieViewSet)
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+
+app_name = "cinema"
+
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
ArsenPidhoretskyi, line 24: No need to include if basepath is empty. Use:
urlpatters = router.urls

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #251, opened 2023-07-04
Solution

xtina999:


installed_packages.txt
ArsenPidhoretskyi, line 1: Why you need this file?

cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = DefaultRouter()
+router.register(r"genres", GenreViewSet)
+router.register(r"actors", ActorViewSet)
+router.register(r"cinema_halls", CinemaHallViewSet)
+router.register(r"movies", MovieViewSet)
+router.register(r"movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
ArsenPidhoretskyi, line 21: No need to include if basepath is not provided. Instead use:
urlpatters = router.urls

ArsenPidhoretskyi (changes requested):

cinema/urls.py
ArsenPidhoretskyi: @xtina999 not fixed

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #253, opened 2023-07-08
solution

IvanGuculyak:


cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
vsmutok, line 21: No need to include if basepath is not provided. Instead use:
urlpatters = router.urls

vsmutok (changes requested):

vsmutok (approved):
Great 👍

PR #265, opened 2023-07-19
Solution

olenaliuby:


cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
+
```
vsmutok, line 23: No need to include if basepath is not provided. Instead use:
urlpatters = router.urls

vsmutok (changes requested):

cinema/urls.py
olenaliuby: Changed. Thank you!

vsmutok (approved):
Good Job 👍

PR #270, opened 2023-07-26
Solution

KirillMelanich:


cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
vsmutok, line 22: No need to include if basepath is not provided. 
Instead use: urlpatterns = router.urls

vsmutok (changes requested):
also remove DB from PR

vsmutok (changes requested):
.

vsmutok (changes requested):
remove DB from PR

vsmutok (approved):
Great 👍

PR #274, opened 2023-07-30
Solution

imelnyk007:


cinema/serializers.py
```diff
@@ -1 +1,76 @@
...
+
+
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Genre
+        fields = ("id", "name")
+
+
+class ActorSerializer(serializers.ModelSerializer):
+    full_name = serializers.CharField(source="*", read_only=True)
```
Oleg-s-s, line 13: nice one, but its bad practice in case actor model will change in the future

Oleg-s-s (approved):

PR #282, opened 2023-07-31
Solution

Gvilion:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return self.first_name + " " + self.last_name
```
Oleg-s-s, line 32: There was no need to change the existing one in def str, but if you are adding something similar it better to use f strings

Oleg-s-s (approved):

PR #295, opened 2023-08-01
Solution

Tarasidze:


cinema/serializers.py
```diff
@@ -1 +1,83 @@
...
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.SlugRelatedField(
+        many=True, slug_field="name", read_only=True
+    )
+    actors = serializers.StringRelatedField(many=True)
+
+    class Meta:
+        model = Movie
+        fields = ("id", "title", "description", "duration", "genres", "actors")
```
YuriiKindrat, line 38: why do u need this ? u are inherit from movieserializer with the same fields and model

cinema/serializers.py
```diff
@@ -1 +1,83 @@
...
+
+
+class MovieDetailSerializer(MovieSerializer):
+    genres = GenreSerializer(many=True)
+    actors = ActorSerializer(many=True)
+
+    class Meta:
+        model = Movie
+        fields = ("id", "title", "description", "duration", "genres", "actors")
+
```
YuriiKindrat, line 48: same

YuriiKindrat (changes requested):

cinema/serializers.py
Tarasidze: I agree with you. At that moment I've decided to write like that for better readable

Oleg-s-s (approved):

PR #298, opened 2023-08-03
My solution

kovaliskoveronika:


cinema/serializers.py
```diff
@@ -1 +1,96 @@
...
+
+class MovieSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = ("id", "title", "description", "duration", "actors", "genres")
+
+
+class MovieListSerializer(MovieSerializer):
+    actors = serializers.SerializerMethodField()
+    genres = serializers.SerializerMethodField()
```
YuriiKindrat, line 50: u can use StringRelatedSerializer here instead

cinema/views.py
```diff
@@ -1 +1,68 @@
...
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all().prefetch_related(
+        "movie",
+        "cinema_hall"
+    )
```
YuriiKindrat, line 60: no need to use `all()` mehtod before joins
also are u sure about prefetch_related for moviesessions ?

YuriiKindrat (changes requested):

cinema/views.py
```diff
@@ -1 +1,68 @@
...
+        return ActorSerializer
+
+
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all().prefetch_related("genres", "actors")
```
LiudmylaKulinchenko, line 45: no need to use all() method before joins

LiudmylaKulinchenko (approved):
Well done 🎈👏

PR #303, opened 2023-08-07
solution

OksanaAdamchuk:


cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+
+router = routers.DefaultRouter()
+router.register("cinema_halls", CinemaHallViewSet, basename="cinema_hall")
+router.register("actors", ActorViewSet)
+router.register("genres", GenreViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+
+urlpatterns = [path("", include(router.urls))]
```
ArsenPidhoretskyi, line 21: No need to include if basepath is not provided. Instead use:
urlpatterns = router.urls

cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+    ActorViewSet,
+    CinemaHallViewSet,
+    GenreViewSet,
+    MovieSessionViewSet,
+    MovieViewSet,
+)
+
+
+router = routers.DefaultRouter()
+router.register("cinema_halls", CinemaHallViewSet, basename="cinema_hall")
```
ArsenPidhoretskyi, line 14: Why you need basename here?

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+    queryset = Movie.objects.prefetch_related("genres", "actors")
+    serializer_class = MovieSerializer
+
+    def get_serializer_class(
+        self,
+    ) -> (
+        type[MovieDetailSerializer]
+        | type[MovieListSerializer]
+        | type[MovieSerializer]
+    ):
```
ArsenPidhoretskyi, line 42:
```suggestion
    ) -> type[Serializer]:
```

ArsenPidhoretskyi: The same fix is in other places. No need to write all possible serializers. Use parent class

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #305, opened 2023-08-12
'Solution'

svitlana-savikhina:


YukihiroSM (changes requested):
Fix tests, please. 
Also, db.sqlite3 shouldn't be in the repo

EdAlekseiev (approved):
Good job

PR #307, opened 2023-08-15
Solution

Oleksiy-Lyashenko:


cinema/views.py
```diff
@@ -1 +1,90 @@
...
+
+
+class GenreViewSet(viewsets.ModelViewSet):
+    queryset = Genre.objects.all()
+    serializer_class = GenreSerializer
+
+
+class ActorViewSet(viewsets.ModelViewSet):
+    queryset = Actor.objects.all()
+    serializer_model = ActorSerializer
```
pavlejviki, line 31: the standard way to specify the serializer for a view or viewset is through the serializer_class attribute, not serializer_model.

cinema/views.py
```diff
@@ -1 +1,90 @@
...
+
+    def get_serializer_class(self):
+        return ActorSerializer
```
pavlejviki, line 34: Instead of overriding the get_serializer_class method to return the ActorSerializer, you can define it directly as a class attribute.

cinema/views.py
```diff
@@ -1 +1,90 @@
...
+
+
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_model = CinemaHallSerializer
+
+    def get_serializer_class(self):
+        return CinemaHallSerializer
```
pavlejviki, line 42: See comment above

cinema/views.py
```diff
@@ -1 +1,90 @@
...
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_model = MovieSerializer
```
pavlejviki, line 47: use serializer_class instead

pavlejviki (changes requested):

cinema/views.py
```diff
@@ -1 +1,84 @@
...
+
+
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
```
Dimosphen1, line 41: There is no need to declare `serializer_class` here in case you use the `get_serializer` case and do not call it from there. In this case, it is code duplication, so remove `serializer_class` here.

cinema/views.py
```diff
@@ -1 +1,84 @@
...
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
```
Dimosphen1, line 65: There is no need to declare `serializer_class` here in case you use the `get_serializer` case and do not call it from there. In this case, it is code duplication, so remove `serializer_class` here.

Dimosphen1 (changes requested):
Several minor changes were requested.

pavlejviki (approved):
Well done!

PR #318, opened 2023-09-13
Solution

Munkoor:


cinema/views.py
```diff
@@ -1 +1,53 @@
...
+    serializer_class = ActorSerializer
+
+
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all().prefetch_related("genres")
```
vasylhnatiuk:
```suggestion
    queryset = Movie.objects.prefetch_related("genres")
```

vasylhnatiuk (changes requested):
delete db.sqilte3 from commit

cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
ArsenPidhoretskyi: No need to include when basepath not provided. Use:
urlpatterns = router.urls

cinema/views.py
ArsenPidhoretskyi: Not fixed
.all() is not needed in this case

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #332, opened 2023-10-10
solution

innyshka:


cinema/serializers.py
```diff
@@ -1 +1,112 @@
...
+        model = MovieSession
+        fields = (
+            "id",
+            "show_time",
+            "movie",
+            "cinema_hall",
+        )
+
+
+class MovieSessionListSerializer(MovieSessionSerializer):
```
eduardhabryd: There is no need to inherit from MovieSessionSerializer if you don't use it's meta class (i have same remark from mentor)

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+    serializer_class = GenreSerializer
+
+
+class ActorViewSet(viewsets.ModelViewSet):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all().prefetch_related("genres", "actors")
```
eduardhabryd: you should remove `.all()` part here

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all().select_related(
```
eduardhabryd: and here

eduardhabryd (changes requested):

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.prefetch_related("genres", "actors")
+    serializer_class = MovieSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
```

Serhii-Leonenko: the same in other similar cases

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

PR #333, opened 2023-10-11
'Solution'

lgSilay:


nyanplague (approved):

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+
+    def get_serializer_class(self):
+
+        if self.action == "list":
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
```
Serhii-Leonenko: return self.serializer_class, the same in all similar cases

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

PR #334, opened 2023-10-11
Solution

nyanplague:


cinema/serializers.py
```diff
@@ -1 +1,76 @@
...
+        )
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
+    movie = MovieListSerializer(read_only=True)
+    cinema_hall = CinemaHallSerializer(read_only=True)
+
+    class Meta:
+        model = MovieSession
+        fields = ("id", "show_time", "movie", "cinema_hall")
```
MarianKovalyshyn: I think this class Meta is redundant because you inherit from MovieSessionSerializer and it already has this class

MarianKovalyshyn (approved):

cinema/serializers.py
```diff
@@ -1 +1,76 @@
...
+        fields = (
+            "id",
+            "show_time",
+            "movie_title",
+            "cinema_hall_name",
+            "cinema_hall_capacity",
+        )
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
```
Serhii-Leonenko, line 70: Do you really need inheritance here from MovieSessionSerializer ?)

cinema/views.py
```diff
@@ -1 +1,67 @@
...
+            queryset = queryset.prefetch_related("genres", "actors")
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        if self.action == "list":
+            return MovieListSerializer
+
+        return MovieSerializer
```

the same in all other similar cases

db.sqlite3
Serhii-Leonenko, line 1: redundant file

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

PR #335, opened 2023-10-11
Implement serializers and views for cinema project

MarianKovalyshyn:


CaCtUs1k (approved):

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return self.first_name + " " + self.last_name
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
Serhii-Leonenko: DRY

cinema/views.py
```diff
@@ -1 +1,61 @@
...
+    ) -> MovieSerializer | MovieListSerializer | MovieDetailSerializer:
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.prefetch_related("cinema_hall", "movie")
```
Serhii-Leonenko: are you sure about prefetch_related here?

cinema/views.py
```diff
@@ -1 +1,61 @@
...
+    ) -> (
+        MovieSessionSerializer
+        | MovieSessionListSerializer
+        | MovieSessionDetailSerializer
+    ):
+        if self.action == "list":
+            return MovieSessionListSerializer
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+        return MovieSessionSerializer
```

the same in all other similar caes

Serhii-Leonenko (changes requested):

cinema/models.py
MarianKovalyshyn: I know that right now they are identical, but even in video Danylo said it's not good to rely on __str__ because it can be changed at any moment so that's why I decided to use "@property" here.

Serhii-Leonenko: rely in str to full_name

Serhii-Leonenko (approved):

PR #351, opened 2023-10-15
Solution

anna-lybid:


cinema/views.py
```diff
@@ -1 +1,69 @@
...
+        if self.action == "list":
+            queryset = queryset.prefetch_related("genres", "actors")
+        return queryset
+
+    def get_serializer_class(self) -> Type[Serializer]:
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
```

Serhii-Leonenko: the same in other similar cases

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

PR #361, opened 2023-10-18
Implemented API Serializer Task for Cinema APP

goldenuni:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return self.first_name + " " + self.last_name
```
EdAlekseiev, line 32: You can use f-string here instead of concatenation

EdAlekseiev (approved):
Good job

cinema/models.py
goldenuni: @EdAlekseiev I choose a concatenation, cause the original __str__ method is written using concatenation. But the f-strings are more preferable, right?

PR #365, opened 2023-11-02
Solution

Veinmax:


cinema/urls.py
```diff
@@ -1 +1,22 @@
...
+router = routers.DefaultRouter()
+router.register("movies", MovieViewSet)
+router.register("actors", ActorViewSet)
+router.register("genres", GenreViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi: No need to include when basepath not provided instead use:
urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #367, opened 2023-11-07
Develop

markus2812:


vasylhnatiuk (changes requested):
fix tests

cinema/urls.py
```diff
@@ -1 +1,42 @@
...
+)
+router.register(
+    "movie_sessions",
+    MovieSessionViewSet,
+    basename="movie-sessions"
+)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi: No need to include if basepath is not provided. Use urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (changes requested):
delete db.sqlite3 from commit

vasylhnatiuk (approved):

PR #368, opened 2023-11-09
Solution

lubart2517:


cinema/serializers.py
```diff
@@ -1 +1,88 @@
-# write serializers here
+from rest_framework import serializers
```
viktoria-rybenchuk, line 1: You should put a blank line between each group of imports.

cinema/serializers.py
```diff
@@ -1 +1,88 @@
...
+from .models import (
+    CinemaHall,
+    Movie,
+    MovieSession,
+    Actor,
+    Genre
+)
```
viktoria-rybenchuk: Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) Read https://peps.python.org/pep-0008/#imports

cinema/views.py
```diff
@@ -1 +1,74 @@
...
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
```
viktoria-rybenchuk: return self.serializer_class. Fix in all places

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #373, opened 2023-11-20
Implement serializers and views for the models

KolBohdan:


cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+
+urlpatterns = [
+    path("", include(router.urls)),
```
pavlejviki: No need to use include when basepath is not provided. 
`urlpatterns = router.urls`

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #379, opened 2023-11-29
Solution

vshvanska:


cinema/models.py
```diff
@@ -30,6 +30,10 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    @property
+    def full_name(self) -> str:
+        return self.first_name + " " + self.last_name
```
mykolamateichuk, line 35: you can use `return str(self)` here

mykolamateichuk (changes requested):

cinema/models.py
vshvanska: I could, but that would make my result dependent on str. What if someone wanted str to also output id? This is also the reason why we  prefer to use slugrelatedfield instead of stringrelatedfield

cinema/serializers.py
```diff
@@ -1 +1,71 @@
...
+from rest_framework import serializers
+
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
+
+
+class CinemaHallSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CinemaHall
+        fields = ("id", "name", "rows", "seats_in_row", "capacity")
+        read_only_fields = ("id",)
```
LiudmylaKulinchenko: No need to specify id is read-only field in your Serializers, it is read-only by default

cinema/serializers.py
```diff
@@ -1 +1,71 @@
...
+    genres = GenreSerializer(many=True, read_only=True)
+    actors = ActorSerializer(many=True, read_only=True)
+
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.SlugRelatedField(many=True,
+                                          read_only=True,
+                                          slug_field="name")
+    actors = serializers.SlugRelatedField(many=True, read_only=True,
+                                          slug_field="full_name")
```
LiudmylaKulinchenko: Fix formatting, you have it different for genres and actors here. Try to remain consistent code style

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great job 🐣

PR #384, opened 2023-11-30
Solution

Daniil-Pankieiev:


RabenkoYevhenii (approved):
Seems good to me

cinema/views.py
```diff
@@ -1 +1,69 @@
...
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
```
e.g.
```python
if self.action in ...:
    return self.queryset....
return self.queryset
```

cinema/views.py
```diff
@@ -1 +1,69 @@
...
+            return MovieRetrieveSerializer
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
```

Y-Havryliv (changes requested):

cinema/views.py
Daniil-Pankieiev: Fixed

LiudmylaKulinchenko (approved):
Good job ☃

PR #387, opened 2023-11-30
Solution

D1st3f:


SevKrok (approved):
Nice

cinema/serializers.py
```diff
@@ -1 +1,79 @@
...
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Genre
+        fields = "__all__"
+
+
+class ActorSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Actor
+        fields = ("first_name", "last_name", "full_name")
```
Y-Havryliv: 2. At the same time movie detail endpoint should provide complete information about the genres and actors.
from README. We will not see actors' ids.

cinema/views.py
```diff
@@ -1 +1,63 @@
...
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self) -> queryset:
+        queryset = self.queryset
```
e.g.
```python
if self.action in ...:
    return self.queryset....
return self.queryset
```

cinema/views.py
```diff
@@ -1 +1,63 @@
...
+            return MovieDetailSerializer
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
+
+    def get_queryset(self) -> queryset:
+        queryset = self.queryset
```

Y-Havryliv (changes requested):

Y-Havryliv (approved):
gj!

PR #389, opened 2023-11-30
sol

Familenko:


cinema/views.py
```diff
@@ -1 +1,67 @@
...
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+
+    def get_queryset(self):
+        queryset = self.queryset
+        if self.action in ("list", "retrieve"):
+            queryset = Movie.objects.prefetch_related("actors", "genres")
```
Naz-iv, line 37: I think you will make two requests to SQL server here since you use Movie manager in line 32 and line 37 ( I may be wrong since querysets are lazy mfs) .  Instead you could use queryset you created in line 35 since it has queryser there already

cinema/serializers.py
```diff
@@ -1 +1,82 @@
...
+    genres = GenreSerializer(many=True, read_only=True)
+    actors = ActorSerializer(many=True, read_only=True)
+
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.StringRelatedField(
+        many=True,
+        read_only=True,
+    )
+    actors = serializers.StringRelatedField(
```
Naz-iv, line 44: maybe it's better to use SlugRelatedField this way your output won't change if someone changes __str__ method

Naz-iv (approved):
looking good

cinema/views.py
Familenko: First query set will be used for method POST PUT etc, where it is not required to have field actors and genre. Another queryset definitely will make 3 request, but it is fine, it fixed amount

cinema/serializers.py
Familenko: Maybe, but I planed not to make property and just retrieve info from __str__ to avoid modification of models completely, but unfortunately test was not able to pass without this field

PR #391, opened 2023-11-30
Develop

Naz-iv:


cinema/views.py
```diff
@@ -1 +1,75 @@
...
+    MovieSessionListSerializer,
+    MovieSessionDetailSerializer,
+    MovieDetailSerializer
+)
+
+
+class MovieViewSet(ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
```
Familenko, line 26: You may delete this row ```serializer_class = MovieSerializer``` becouse anyway your serializer_class updated in ```def get_serializer_class(self):```

Familenko (approved):

petrykivd (approved):

cinema/serializers.py
```diff
@@ -1 +1,105 @@
...
+        fields = (
+            "id",
+            "show_time",
+            "movie",
+            "cinema_hall",
+        )
+
+
+class MovieSessionListSerializer(MovieSessionSerializer):
+    movie_title = serializers.CharField(source="movie.title",
```
Y-Havryliv: just an advice, use the same style as below

cinema/serializers.py
```diff
@@ -1 +1,105 @@
...
+from rest_framework import serializers
+from rest_framework.serializers import ModelSerializer
+
+from cinema.models import Genre, Actor, Movie, MovieSession, CinemaHall
+
+
+class GenreSerializer(ModelSerializer):
+    class Meta:
+        model = Genre
+        fields = "__all__"
```
Y-Havryliv, line 10: you specified field names in all project except this place

Y-Havryliv: keep it one-styled

cinema/views.py
```diff
@@ -1 +1,75 @@
...
+    def get_queryset(self):
+        queryset = self.queryset
```
e.g.
```python
if self.action in ...:
    return self.queryset....
return self.queryset
```

cinema/views.py
```diff
@@ -1 +1,75 @@
...
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieSessionViewSet(ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
```

Y-Havryliv (changes requested):

cinema/serializers.py
Naz-iv: corrected

cinema/serializers.py
Naz-iv: I specified fields in most serializers because task required json objects with specific representation. I kept is as "__all__" for serializers which use only fields that are provide in models.

cinema/views.py
Naz-iv: just followed example from conspectus, corrected now

LiudmylaKulinchenko (approved):
Good job 🎄☃

PR #395, opened 2023-12-01
Solution

Wonsky1:


cinema/models.py
```diff
@@ -30,13 +30,17 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
```
Lyutillis: U can return just str(self) here

cinema/views.py
```diff
@@ -1 +1,55 @@
...
+class GenreViewSet(viewsets.ModelViewSet):
+    queryset = Genre.objects.all()
+    serializer_class = GenreSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all().select_related("cinema_hall")
+    serializer_class = MovieSessionSerializer
+
+    def get_serializer_class(self):
```
Lyutillis: No need to use all(). Also, add movie to select_related.

cinema/views.py
```diff
@@ -1 +1,55 @@
...
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+
+        return MovieSessionSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all().prefetch_related("genres", "actors")
+    serializer_class = MovieSerializer
+
```
Lyutillis: The same here. Remove .all().

Lyutillis (changes requested):

LiudmylaKulinchenko (approved):
Great 🎄

PR #397, opened 2023-12-01
Solution/yaroslav-lysiuk

Quiet-Klirik:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return self.first_name + " " + self.last_name
```
viktoria-rybenchuk: Consider using an `f-string` for better readability and conciseness.

cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
viktoria-rybenchuk: Omit the router when there's no specified basepath. Utilize `urlpatterns = router.urls`  instead.

viktoria-rybenchuk (changes requested):

cinema/models.py
Quiet-Klirik: It is implemented feature by Mate

vasylhnatiuk (approved):

PR #408, opened 2023-12-12
Solution

asdadaversa:


user/migrations/0001_initial.py
```diff
@@ -11,122 +11,34 @@ class Migration(migrations.Migration):
...
+                'verbose_name': 'user',
+                'verbose_name_plural': 'users',
+                'abstract': False,
             },
             managers=[
-                ("objects", django.contrib.auth.models.UserManager()),
+                ('objects', django.contrib.auth.models.UserManager()),
             ],
         ),
     ]
```
Dimosphen1: Do not overwrite or modify the existing migrations, create a new one in case needed.

cinema/views.py
```diff
@@ -1 +1,57 @@
...
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.prefetch_related(
+        "genres"
+    ).prefetch_related("actors")
+
+    serializer_class = MovieSerializer
```
Dimosphen1: There is no need to define the `serializer_class` attribute in case you implement `get_serializer_class` method. Make changes to your code to encounter that.

Dimosphen1 (changes requested):
Several changes were requested.

cinema/urls.py
```diff
@@ -1 +1,25 @@
...
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+
+urlpatterns = [
+    path("", include(router.urls)),
```
pavlejviki: if you don't need to add add a path prefix you can directly assign router.urls to urlpatterns without using include.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #419, opened 2024-01-05
Implement serializers and views for the models: Genre, Actor, CinemaHa…

DenysRaboshchuk:
…ll, Movie, MovieSession

### DIFF
--- a/cinema/models.py
+++ b/cinema/models.py
@@ -30,6 +30,10 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    @property
+    def full_name(self):
+        return self.__str__
+
 
 class Movie(models.Model):
     title = models.CharField(max_length=255)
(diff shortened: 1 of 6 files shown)


cinema/models.py
```diff
@@ -30,6 +30,10 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    @property
+    def full_name(self):
+        return self.__str__
```
NikOneZ1: You don't call the function, so you will return function object

cinema/serializers.py
```diff
@@ -1 +1,98 @@
...
+    Movie,
+    MovieSession,
+)
+
+
+class GenreSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = Genre
+        fields = "__all__"
```
NikOneZ1: GenreSerializer should use only id and name fields

NikOneZ1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #427, opened 2024-01-20
Solution

vkrasnovyd:


cinema/urls.py
```diff
@@ -1 +1,27 @@
...
+router.register("movies", MovieViewSet, basename="movie")
+router.register(
+    "movie_sessions",
+    MovieSessionViewSet,
+    basename="movie-session"
+)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi: No need to include router when basepath is not provided. Use urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #428, opened 2024-01-21
FOR HELP

Shum430:


cinema/urls.py
```diff
@@ -1 +1,45 @@
...
+router.register(
+    "movie_sessions",
+    MovieSessionViewSet,
+    basename="movie_session"
+)
+
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
ArsenPidhoretskyi: No need to include router when basepath is not provided. Use:
urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done

PR #442, opened 2024-02-15
Solution for API serializers

AlimFreight:


cinema/models.py
```diff
@@ -27,7 +27,8 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
-    def __str__(self):
+    @property
```
ArsenPidhoretskyi, line 30: Why do you remove str method?

cinema/serializers.py
```diff
@@ -1 +1,83 @@
...
+
+
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.Genre
+        fields = "__all__"
+
+
+class GenreViewSerializer(GenreSerializer):
+    pass
```
ArsenPidhoretskyi: This class is redundant

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #443, opened 2024-02-16
Solution

Zonda001:


cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = routers.DefaultRouter()
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi: No need to include router when basepath is not provided
Use urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
.

ArsenPidhoretskyi (approved):
Well done!

PR #450, opened 2024-02-22
Solution

Lebwa1337:


cinema/serializers.py
```diff
@@ -1 +1,87 @@
...
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
+
+
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Genre
+        fields = "__all__"
+
+
+class ActorSerializer(serializers.ModelSerializer):
```
viktoria-rybenchuk: import `ModelSerializer` directly from  from  **rest_framework.serializers**

viktoria-rybenchuk: Fix in all places

viktoria-rybenchuk (changes requested):
No changes are required in the initial migration file. Instead, add a new migration file for the latest changes.

Lebwa1337:
Initial migrations always doesn`t work for me. So, in this case i need to run new migrations for make them work

viktoria-rybenchuk (approved):
GJ!

PR #461, opened 2024-03-12
Solution

ArturPoltser:


MaxymChyncha (approved):
LGTM

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
MNDonut, line 35: if you use the same string, we can use one method inside other one

cinema/serializers.py
```diff
@@ -1 +1,90 @@
...
+            "show_time",
+            "movie_title",
+            "cinema_hall_name",
+            "cinema_hall_capacity"
+        )
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
+    movie = MovieListSerializer(many=False, read_only=True)
+    cinema_hall = CinemaHallSerializer(many=False, read_only=True)
```
MNDonut, line 90: many is False by default, so you don't need to pass it

MNDonut (approved):

PR #465, opened 2024-03-13
Solution

kabust:


cinema/serializers.py
```diff
@@ -1 +1,72 @@
-# write serializers here
+from rest_framework import serializers
+
+from cinema.models import Actor, Genre, CinemaHall, Movie, MovieSession
+
+
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
```
aLEKS-e3, line 8: Ihor suggested to leave a blank line between main class and a sub class)

aLEKS-e3 (approved):

cinema/serializers.py
kabust: ty, will fix

MNDonut: @aLEKS-e3 10 points to Gryffindor!

cinema/serializers.py
```diff
@@ -1 +1,72 @@
...
+class CinemaHallSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CinemaHall
+        fields = ("id", "name", "rows", "seats_in_row", "capacity")
+
+
+class MovieSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = ("id", "title", "description", "duration", "genres", "actors")
```
MNDonut: don't forget about 1 empty line

cinema/serializers.py
```diff
@@ -1 +1,72 @@
...
+            "show_time",
+            "movie_title",
+            "cinema_hall_name",
+            "cinema_hall_capacity"
+        )
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
+    movie = MovieListSerializer(many=False, read_only=True)
+    cinema_hall = CinemaHallSerializer(many=False, read_only=True)
```
MNDonut: many parameter is False by default, so you don't need to add it here

cinema/serializers.py
```diff
@@ -1 +1,72 @@
...
+
+
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Genre
+        fields = "__all__"
+
+
+class ActorSerializer(serializers.ModelSerializer):
+    full_name = serializers.CharField(source="__str__", read_only=True)
```
MNDonut: I've never seen such an implementation with "__str__" as source :0. Let's create a property instead because other devs won't be sure that __str__ indeed contains first and last name + it may be changed in the future, but property will be the same

MNDonut (changes requested):

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return self.first_name + " " + self.last_name
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
MNDonut, line 35: here are the same strings so you can use one method in other one

MNDonut (approved):

PR #467, opened 2024-03-13
Solution

pnakongit:


cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi: Use urlpatterns = router.urls when basepath is not provided

ArsenPidhoretskyi (changes requested):

cinema/urls.py
pnakongit: Noted! Thx you!

vasylhnatiuk (approved):

PR #480, opened 2024-03-15
Solution

msymonovych:


cinema/views.py
```diff
@@ -1 +1,71 @@
...
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset.prefetch_related("genres", "actors")
+        return queryset
```
MNDonut: it's better to do prefetch_related in queryset attribute directly instead of overriding get_queryset (it'll be the same, but easier to read)

cinema/views.py
```diff
@@ -1 +1,71 @@
...
+        return self.serializer_class
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset.prefetch_related("movie", "cinema_hall")
+        return queryset
```
MNDonut: let's remove this redundant overrided get_queryset method

MNDonut (changes requested):

MNDonut (approved):

PR #483, opened 2024-03-17
Sol

Dobryk12:


JohnnyWalker010 (approved):

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
MNDonut: if your str and full name are the same, maybe it's better to use one method inside another one?

cinema/serializers.py
```diff
@@ -1 +1,98 @@
...
+            "rows",
+            "seats_in_row"
+        ]
+
+
+class MovieSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = [
+            "id",
```
MNDonut: 1 empty line

cinema/serializers.py
```diff
@@ -1 +1,98 @@
...
+            "show_time",
+            "movie_title",
+            "cinema_hall_name",
+            "cinema_hall_capacity"
+        )
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
+    movie = MovieListSerializer(many=False, read_only=True)
+    cinema_hall = CinemaHallSerializer(many=False, read_only=True)
```
MNDonut: many is False by default, thus you don't need to set it explicitly

MNDonut (changes requested):

cinema/serializers.py
```diff
@@ -94,5 +101,6 @@ class Meta:
 
 
 class MovieSessionDetailSerializer(MovieSessionSerializer):
+
     movie = MovieListSerializer(many=False, read_only=True)
     cinema_hall = CinemaHallSerializer(many=False, read_only=True)
```
MNDonut: remove many parameter, it's False by default

MNDonut (changes requested):

cinema/serializers.py
Dobryk12: > remove many parameter, it's False by default

Oh sorry, some mentors told that it's imortant to writedown it, but I ll change

Serhii-Leonenko (approved):

PR #486, opened 2024-03-17
Solution

Serejka21:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
kstorozhenko, line 36: In my opinion in the __str__ method you can just return the full name

kstorozhenko (approved):

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
...
```
MNDonut: DRY

cinema/serializers.py
```diff
@@ -1 +1,102 @@
...
+            "show_time",
+            "movie_title",
+            "cinema_hall_name",
+            "cinema_hall_capacity"
+        )
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
+    movie = MovieListSerializer(many=False, read_only=True)
+    cinema_hall = CinemaHallSerializer(many=False, read_only=True)
```
MNDonut, line 102: many is False by default, thus no need to pass it

cinema/views.py
```diff
@@ -1 +1,71 @@
...
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
```
MNDonut: self.serializer_class

cinema/views.py
```diff
@@ -1 +1,71 @@
...
+    serializer_class = MovieSessionSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+
+        return MovieSessionSerializer
```
MNDonut: self.serializer_class

cinema/views.py
```diff
@@ -1 +1,71 @@
...
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+
+        if self.action in ("list", "retrieve"):
+            return queryset.prefetch_related("actors", "genres")
```
MNDonut: I think you don't need this check. For example when you do PUT, you update instance and it calls self.get_object() that uses self.get_queryset() to get object. It means when PUT returns object, in your case it won't use prefetch_related

MNDonut (changes requested):

cinema/views.py
```diff
@@ -1 +1,71 @@
...
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+
+        if self.action in ("list", "retrieve"):
+            return queryset
+
```
MNDonut: what I meant here is that for other actions you need prefetch_related as well, thus you can remove get_queryset and just use prefetch_related in queryset attribute

MNDonut (changes requested):

MNDonut (approved):

PR #490, opened 2024-03-19
Solution

NPovashevych:


cinema/urls.py
```diff
@@ -1 +1,28 @@
...
+router.register("cinema_halls", CinemaHallViewSet, basename="cinema-halls")
+router.register("movies", MovieViewSet, basename="movies")
+router.register(
+    "movie_sessions",
+    MovieSessionViewSet,
+    basename="movies-sessions")
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi: No need to include router, instead use:
urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #495, opened 2024-03-20
Solution

isikach:


cinema/serializers.py
```diff
@@ -1 +1,87 @@
...
+
+class GenreSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = Genre
+        fields = "__all__"
+
+
+class ActorSerializer(serializers.ModelSerializer):
+    full_name = serializers.CharField(source="__str__", read_only=True)
```
MNDonut: I don't recommend to use __str__ as source because it is often changed, let's create a property for this purpose

cinema/serializers.py
```diff
@@ -1 +1,87 @@
...
+
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.SlugRelatedField(
+        many=True,
+        read_only=True,
+        slug_field="name"
+    )
+    actors = serializers.StringRelatedField(many=True)
+
```
MNDonut: read the requirements please, this field will give you another output

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
```
MNDonut: self.serializer_class

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+
+        if self.action in ("list", "retrieve"):
+            return queryset.prefetch_related("actors", "genres")
```
MNDonut: do you need this check? in other operations you works with a single object, thus you are free to override queryset attribute instead of get_queryset

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+    serializer_class = MovieSessionSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+
+        return MovieSessionSerializer
```
MNDonut: self.serializer_class

MNDonut (changes requested):

cinema/views.py
```diff
@@ -33,22 +33,14 @@ class MovieViewSet(viewsets.ModelViewSet):
     queryset = Movie.objects.all()
```
Serhii-Leonenko: You should use prefetch_related here, which you used in the get_queryset

cinema/models.py
```diff
@@ -30,6 +30,10 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    @property
+    def actors_full_name(self):
+        return str(self)
```
Serhii-Leonenko: Use property in the str, because str could be changed and the logic of the property field will be changed as well

cinema/serializers.py
```diff
@@ -11,7 +11,10 @@ class Meta:
 
 
 class ActorSerializer(serializers.ModelSerializer):
-    full_name = serializers.CharField(source="__str__", read_only=True)
+    full_name = serializers.CharField(
+        source="actors_full_name",
+        read_only=True
+    )
```
Serhii-Leonenko: you can change the name of the property to full_name, in a result you don't need to define this field here with source

Serhii-Leonenko (changes requested):

Y-Havryliv (approved):
gj!

PR #504, opened 2024-04-05
solution for py-api-serializers

oleksiimuhyr:


vasylhnatiuk (changes requested):
delete db.sqlite3 from pull request

vasylhnatiuk (changes requested):
use rm cached to delete db.sqlite3 https://timmousk.com/blog/git-rm-cached/

EdAlekseiev (approved):

PR #506, opened 2024-04-15
Solution

NataliiaKucher:


EdAlekseiev (changes requested):
Fix tests, please

cinema/serializers.py
```diff
@@ -1 +1,145 @@
...
+        read_only=True,
+        slug_field="name"
+    )
+    actors = serializers.SerializerMethodField()
+
+    def get_actors(self, obj):
+        return [
+            f"{actor.first_name} {actor.last_name}"
+            for actor in obj.actors.all()
+        ]
```
EdAlekseiev: Why is `actors`  field `SerializerMethodField`? You should use  serializers.SlugRelatedField. Also, you have to add `full_name` property to `Actor`
```suggestion
    actors = serializers.SlugRelatedField(
        many=True,
        read_only=True,
        slug_field="full_name"
    )
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #511, opened 2024-04-19
Solution

mkuchuman:


cinema/urls.py
```diff
@@ -1 +1,27 @@
...
+                GenreViewSet, basename="genres")
+router.register("movies",
+                MovieViewSet, basename="movies")
+router.register("movie_sessions",
+                MovieSessionViewSet, basename="movie_sessions")
+
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi: When basepath is empty no need to include, instead use:
urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

cinema/urls.py
ArsenPidhoretskyi: This issue not fixed

ArsenPidhoretskyi (changes requested):

mkuchuman:
Now it is?

mkuchuman:
I forgot to push changes

ArsenPidhoretskyi (approved):
Well done!

PR #512, opened 2024-04-25
Solution

mwellick:


db.sqlite3
EdAlekseiev, line 1: Delete db.sqlite3 and add it to .gitignore

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #517, opened 2024-05-02
solution

SysoevDmitro:


db.sqlite3
EdAlekseiev, line 1: You shouldn't store db files in repository, because these files can be large. Add sqlite3 file to .gitignore and delete it from repository.

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #524, opened 2024-05-15
Solution

twonumberfortyfives:


cinema/urls.py
```diff
@@ -1 +1,22 @@
...
+
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
```
ArsenPidhoretskyi: No need to use include when basepath is empty. Use:
urlpatterns = router.urls

cinema/views.py
```diff
@@ -1 +1,76 @@
...
+        if self.action == "list":
+            return MovieSessionListSerializer
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+        return MovieSessionSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+        if self.action == "list":
+            queryset = queryset.select_related()
```
ArsenPidhoretskyi: Specify what should select

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):

PR #531, opened 2024-05-22
Solution

donbas777:


cinema/urls.py
```diff
@@ -1 +1,21 @@
...
+
+router = routers.DefaultRouter()
+router.register("actors", ActorViewSet)
+router.register("genres", GenreViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+
+urlpatterns = [] + router.urls
```
Dimosphen1, line 21: There is no need to include `[] +` part here, you could assign `router.urls` directly to `urlpatterns` instead in the future

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #532, opened 2024-05-22
Solution

Artem-user-19:


pavlopro (changes requested):
Do not modify migration files
Do not modify any test files
Do not delete admin.py
Remove db.sqlite3
Include sqlite3 into git ignore

pavlopro (changes requested):
Do not modify migration files
Do not modify any test files
Do not delete admin.py

PR #535, opened 2024-05-23
Solution

Artem-user-19:


cinema/views.py
```diff
@@ -1 +1,50 @@
-# write views here
+from rest_framework import viewsets
+from .models import Genre, Actor, CinemaHall, Movie, MovieSession
+from .serializers import (
```
viktoria-rybenchuk: use absolute imports instead of relative

cinema/views.py
```diff
@@ -1 +1,50 @@
-# write views here
+from rest_framework import viewsets
+from .models import Genre, Actor, CinemaHall, Movie, MovieSession
```
viktoria-rybenchuk: add a blank line between different group of imports

viktoria-rybenchuk (changes requested):
Remove db from the PR

EdAlekseiev (approved):
Good job

PR #541, opened 2024-05-26
Solution to serializers task

panicua:
I uninstalled debug toolbar after testing queries, you can trust me :)


cinema/views.py
```diff
@@ -1 +1,66 @@
...
+class MovieViewSet(viewsets.ModelViewSet):
+    serializer_class = MovieSerializer
+    queryset = Movie.objects.all()
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
```
MNDonut: return self.serializer_class

cinema/views.py
```diff
@@ -1 +1,66 @@
...
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    serializer_class = MovieSessionSerializer
+    queryset = MovieSession.objects.all()
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+        return MovieSessionSerializer
```
MNDonut: return self.serializer_class

MNDonut (approved):

cinema/views.py
panicua: > return self.serializer_class

I changed it, but isn't it more explicit the way I did? especially when other returns are DetailSerializer, ListSerializer

PR #549, opened 2024-05-27
Solution

SashaKisliy:


cinema/views.py
```diff
@@ -1 +1,60 @@
...
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+        if self.action == "retrieve":
+            return MovieSessionRetrieveSerializer
+        return MovieSessionSerializer
+
+    def get_queryset(self):
+        if self.action in ["list", "retrieve"]:
+            return MovieSession.objects.select_related()
```
MNDonut: What if we do PATCH or PUT or DELETE? get_queryset will return None

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Well done 💪

PR #551, opened 2024-05-27
Solution

Chub-andrew:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
MNDonut: return self.full_name?

cinema/serializers.py
```diff
@@ -1 +1,90 @@
-# write serializers here
+from rest_framework import serializers
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
```
MNDonut: 1 blank line between import groups

cinema/serializers.py
```diff
@@ -1 +1,90 @@
...
+    def get_cinema_hall_name(self, obj):
+        return obj.cinema_hall.name
+
+    def get_cinema_hall_capacity(self, obj):
+        return obj.cinema_hall.capacity
+
+
+class MovieListSerializer(serializers.ModelSerializer):
+    genres = serializers.StringRelatedField(many=True)
+    actors = serializers.SerializerMethodField()
```
MNDonut, line 62: why not using the same way? StringRelatedField for both fields or SerializerMethodField for both fields

MNDonut (changes requested):

cinema/serializers.py
Chub-andrew: in the first case I use a string, and in the second method get_actors

cinema/models.py
Chub-andrew: here we are talking about code duplication??
method str duplicated with method full_name

YuriiKindrat: yeah, so u could use full name property in ur str method

YuriiKindrat (changes requested):
remoe db from ur PR

fsocie7y (approved):
GJ 👍

PR #558, opened 2024-05-31
Solution

skeletronys:


cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = DefaultRouter()
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("genres", GenreViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
Dimosphen1, line 21: In case `urlpatterns` consist of only router urls, you could assign `router.urls` to `urlpatterns` directly. Follow this approach in the future

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #559, opened 2024-05-31
Solution

KravchenkoSerhiii:


cinema/models.py
```diff
@@ -27,6 +27,9 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
Dimosphen1: You could return `self.full_name` here instead of the current implementation

cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
Dimosphen1: In case `urlpatterns` consist of only router urls, you could assign `router.urls` to `urlpatterns` directly

cinema/views.py
```diff
@@ -1 +1,69 @@
...
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieRetrieveSerializer
+        return MovieSerializer
```
Dimosphen1: Return `self.serializer_class` here and in the other views instead

cinema/views.py
```diff
@@ -1 +1,69 @@
...
+
+    def get_queryset(self):
+        queryset = self.queryset
+        if self.action in ("list", "retrieve"):
+            return queryset.prefetch_related("genres")
+        return queryset
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
```
Dimosphen1, line 57: Define `serializer_class` class attribute here to remain consistency between views

Dimosphen1 (changes requested):
Several changes were requested

cinema/views.py
pavlopro: This was not resolved. According to the suggestion, you should return self.serializer_class instead of MovieSerializer. Please do this in the other views as well.

pavlopro (changes requested):

vasylhnatiuk (approved):

PR #581, opened 2024-07-04
Solution

frontbastard:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
EdAlekseiev: Use property  `full_name` here, please

db.sqlite3
EdAlekseiev, line 1: You should add database file to .gitignore and delete it from repository

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #586, opened 2024-07-13
Solution

romahrynchyshyn:


cinema/models.py
```diff
@@ -30,28 +36,45 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
```
vsmutok: Use property  `full_name` here, please

vsmutok (changes requested):

cinema/models.py
lmidzhak: you mean to delete str method and just use lines 39-41?

vasylhnatiuk (approved):

PR #602, opened 2024-07-29
solution

necobra:


cinema/urls.py
```diff
@@ -1 +1,19 @@
...
+
+router.register("cinema_halls", views.CinemaHallViewSet)
+router.register("genres", views.GenreViewSet)
+router.register("actors", views.ActorViewSet)
+router.register("movies", views.MovieViewSet)
+router.register("movie_sessions", views.MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
urlpatterns = router.urls

cinema/views.py
```diff
@@ -1 +1,42 @@
...
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = models.Movie.objects.prefetch_related("genres", "actors")
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return serializers.MovieListSerializer
+        if self.action == "retrieve":
+            return serializers.MovieRetrieveSerializer
+        return serializers.MovieSerializer
```
DevSpaciX: Could you return here self.serializer_class ?

cinema/views.py
```diff
@@ -1 +1,42 @@
...
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = (models.MovieSession.objects
+                .select_related("movie", "cinema_hall"))
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return serializers.MovieSessionListSerializer
+        if self.action == "retrieve":
+            return serializers.MovieSessionRetrieveSerializer
+        return serializers.MovieSessionSerializer
```
DevSpaciX: Could you return here self.serializer_class ?

DevSpaciX (changes requested):
PLS remove DB file from PR

Arsen-hrynevych (changes requested):
Thanks. Please format the code using the tool because some parts of the code are difficult to read.

necobra:
> Thanks. Please format the code using the tool because some parts of the code are difficult to read.

after using tools for formating you usually have problems with flake8

Arsen-hrynevych (approved):
Thanks

PR #607, opened 2024-07-29
Solution_v1

BileichukIvan:


DevSpaciX (changes requested):
pls remove db.sqlite3 from PR

DevSpaciX (changes requested):
image
I see you added it in git ignore but anyway it is still in PR

cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
urlpatterns = router.urls

cinema/views.py
```diff
@@ -1 +1,69 @@
...
+        if self.action in ("list", "retrieve"):
+            return queryset.prefetch_related("genres", "actors")
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieRetrieveSerializer
+        return MovieSerializer
```

cinema/views.py
```diff
@@ -1 +1,69 @@
...
+        if self.action in ("list", "retrieve"):
+            return queryset.select_related("movie", "cinema_hall")
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+        elif self.action == "retrieve":
+            return MovieSessionRetrieveSerializer
+        return MovieSessionSerializer
```
DevSpaciX: Same here )

DevSpaciX (changes requested):

DevSpaciX (approved):
well done

PR #608, opened 2024-07-29
complete py-api-serializers v1

BornToLivee:


cinema/urls.py
```diff
@@ -1 +1,22 @@
...
+router = DefaultRouter()
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("genres", GenreViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("movies", MovieViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
urlpatterns = router.urls

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #610, opened 2024-07-29
Solution v.1

SemenetsSergii:
image
image
image
image
image
image

### DIFF
--- a/cinema/urls.py
+++ b/cinema/urls.py
@@ -20,7 +20,7 @@ router.register(
     basename="movie_session"
 )
 
-urlpatterns = [path("", include(router.urls))]
+urlpatterns = router.urls
 
 
 app_name = "cinema"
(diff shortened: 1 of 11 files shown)


cinema/urls.py
```diff
@@ -1 +1,26 @@
...
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register(
+    "movie_sessions",
+    MovieSessionViewSet,
+    basename="movie_session"
+)
+
+urlpatterns = [path("", include(router.urls))]
+
```
urlpatterns = router.urls

DevSpaciX (changes requested):

cinema/urls.py
SemenetsSergii: Uploaded to my brain ;)

DevSpaciX (approved):

PR #613, opened 2024-07-29
Solution

b3v3kt0r:


cinema/urls.py
```diff
@@ -1 +1,21 @@
...
+app_name = "cinema"
+
+router = routers.DefaultRouter()
+router.register("actors", ActorViewSet)
+router.register("genres", GenreViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+
+urlpatterns = [path("", include(router.urls))]
```
urlpatterns = router.urls

cinema/views.py
```diff
@@ -1 +1,70 @@
...
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieRetrieveSerializer
+        return MovieSerializer
```

cinema/views.py
```diff
@@ -1 +1,70 @@
...
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+        elif self.action == "retrieve":
+            return MovieSessionRetrieveSerializer
+        return MovieSessionSerializer
```
DevSpaciX: Same here

cinema/views.py
```diff
@@ -1 +1,70 @@
...
+
+    def get_queryset(self):
+        queryset = self.queryset
+        if self.action in ["list", "retrieve"]:
+            return queryset.select_related()
```
DevSpaciX: select_related() has no effect , did you forget to add necessery models here ?

DevSpaciX (changes requested):

cinema/views.py
DevSpaciX: @b3v3kt0r Did you forget to fix it ?

DevSpaciX (changes requested):

cinema/views.py
b3v3kt0r: @DevSpaciX  yup, sorry

DevSpaciX (approved):

PR #615, opened 2024-07-30
sol

vladyslavchichkan:


cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
urlpatterns = router.urls

cinema/views.py
```diff
@@ -1 +1,70 @@
...
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+        elif self.action == "retrieve":
+            return MovieSessionRetrieveSerializer
+        return MovieSessionSerializer
```

cinema/views.py
```diff
@@ -1 +1,70 @@
...
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieRetrieveSerializer
+        return MovieSerializer
```

DevSpaciX (changes requested):

cinema/views.py
DevSpaciX: Oh , and delete pls db.sqlite3 file from PR

cinema/views.py
Unlie9: please explain why i need to return self.serializer_class here

Arsen-hrynevych (approved):
Thanks for your effort

PR #617, opened 2024-07-30
'Solution'

Irina17191:


cinema/urls.py
```diff
@@ -1 +1,26 @@
...
+router.register("genres", GenreViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
urlpatterns = router.urls

cinema/views.py
```diff
@@ -1 +1,73 @@
...
+    serializer_class = MovieSessionListSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+        elif self.action == "retrieve":
+            return MovieSessionRetrieveSerializer
+        elif self.action == "create":
+            return MovieSessionSerializer
+        return MovieSessionListSerializer
```
DevSpaciX: Do not rewrite serializer_class at the start , do it only in list or retrieve actions . And at the end you can just return self.serializer_class

DevSpaciX (changes requested):

cinema/views.py
```diff
@@ -1 +1,73 @@
...
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieRetrieveSerializer
+        return MovieSerializer
```

DevSpaciX (changes requested):

maxgerman (approved):
Good job

PR #619, opened 2024-08-03
"Solution"

MaglorFeanorson:


cinema/serializers.py
```diff
@@ -1 +1,78 @@
-# write serializers here
+from rest_framework import serializers
+
+from cinema.models import Actor, CinemaHall, Genre, Movie, MovieSession
```
Arsen-hrynevych: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

cinema/views.py
```diff
@@ -1 +1,63 @@
-# write views here
+from rest_framework.viewsets import ModelViewSet
+from cinema.models import Actor, CinemaHall, Genre, Movie, MovieSession
+from cinema.serializers import (
```
Arsen-hrynevych: Imports should be grouped and separated by a blank line according to standard library, external libraries, and custom imports. This practice enhances both code readability and maintainability. 

Arsen-hrynevych (approved):
Good job! I really appreciate your effort and attention to detail. Thank you for your hard work.

PR #628, opened 2024-08-16
Solution

IraMladanovych02:


cinema/migrations/0004_alter_genre_name.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 4.0.4 on 2022-06-16 08:42
+# Generated by Django 4.0.4 on 2024-08-16 16:30
```
pavlopro: Please do not modify migration files

pavlopro (approved):
Good Job!

cinema/views.py
```diff
@@ -1 +1,70 @@
...
+    serializer_class = MovieSessionSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+
+        return MovieSessionSerializer
```
viktoria-rybenchuk:
```suggestion
        return self.serializer_class
```

cinema/views.py
```diff
@@ -1 +1,70 @@
...
+    serializer_class = MovieSerializer
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+
+        return MovieSerializer
```
viktoria-rybenchuk: the same here

viktoria-rybenchuk (changes requested):

cinema/views.py
IraMladanovych02: Thanks for you comment! Could you please explain the difference? Because in lessons I found that we should use MovieSessionSerializer?

viktoria-rybenchuk (approved):
Well done!

PR #632, opened 2024-08-30
Solution

mykytaso:


cinema/urls.py
```diff
@@ -1 +1,28 @@
...
+router.register("movies", MovieViewSet, basename="movie")
+router.register(
+    "movie_sessions",
+    MovieSessionViewSet,
+    basename="movie_session"
+)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
vsmutok: If your path equal to empty string, no reason to specify it if router has been created. Use `urlpatterns = router.urls` instead

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #635, opened 2024-09-08
Implement solution

dimak20:


cinema/urls.py
```diff
@@ -1 +1,43 @@
...
+)
+router.register(
+    "movie_sessions",
+    MovieSessionViewSet,
+    basename="movie_session"
+)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
Arsen-hrynevych, line 43: [optional] You could simplify it to urlpatterns = router.urls

Arsen-hrynevych (approved):
Thanks

cinema/urls.py
dimak20: Did not know about that. Thanks!

PR #637, opened 2024-09-11
Implement API

anton-bilous:


cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+router = routers.SimpleRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
Arsen-hrynevych, line 24: [optional] You could change it => urlpatterns = router.urls

Arsen-hrynevych (approved):
Overall, your work is good, but your commits need improvement. They are not descriptive, and you shouldn't commit every small change. Your commits should be atomic, focusing on specific parts of what you've completed.

PR #645, opened 2024-09-26
solution

BastovOleksandr:


cinema/serializers.py
```diff
@@ -1 +1,72 @@
-# write serializers here
+from rest_framework import serializers
+
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
+
+
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Genre
+        fields = "__all__"
```
MasakDirt: Maybe it is better to add all usable fields here, instead of `__all__`, then it will be more readable

cinema/views.py
```diff
@@ -1 +1,66 @@
...
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+
+    def get_queryset(self):
+        queryset = self.queryset
+        if self.action in ("list", "retrieve"):
+            queryset = self.queryset.select_related("movie", "cinema_hall")
+            if self.action == "retrieve":
```
MasakDirt, line 55: You twice check here if action is retrieve, you can just do this separately like in `get_serializer_class` methods

MasakDirt (changes requested):

MasakDirt (approved):

MasakDirt (approved):

cinema/urls.py
```diff
@@ -1 +1,21 @@
...
+)
+
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [path("", include(router.urls))]
```
Arsen-hrynevych, line 19: You could simplify it to: `urlpatterns = router.urls`.

cinema_service/settings.py
```diff
@@ -42,10 +47,12 @@
     "rest_framework",
     "cinema",
     "user",
+    # "debug_toolbar",
 ]
 
 MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
+    # "debug_toolbar.middleware.DebugToolbarMiddleware",
```
Arsen-hrynevych, line 55: ?

cinema_service/urls.py
```diff
@@ -1,6 +1,8 @@
 from django.contrib import admin
-from django.urls import path
+from django.urls import path, include
 
 urlpatterns = [
     path("admin/", admin.site.urls),
+    path("api/cinema/", include("cinema.urls", namespace="cinema")),
+    # path("__debug__/", include("debug_toolbar.urls")),
```
Arsen-hrynevych, line 7: Commented-out code shouldn't be included in the PR.

cinema/serializers.py
```diff
@@ -1 +1,73 @@
-# write serializers here
+from rest_framework import serializers
+
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

Arsen-hrynevych (approved):
Thanks

PR #647, opened 2024-09-28
Solution

dartomOOv:


cinema/serializers.py
```diff
@@ -1 +1,83 @@
-# write serializers here
+from rest_framework import serializers
+
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
```
dimak20: Maybe It's better to split improts down one by one

cinema/serializers.py
```diff
@@ -1 +1,83 @@
...
+        fields = ["id", "title", "description", "duration", "genres", "actors"]
+
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.SlugRelatedField(
+        many=True,
+        read_only=True,
+        slug_field="name"
+    )
+    actors = serializers.SerializerMethodField()
```
dimak20, line 49: U can use just some related method with many=True instead of this func but It's also OK

dimak20 (approved):

LiudmylaKulinchenko (approved):
Well done 🌸

PR #650, opened 2024-09-29
first commit

panichqa:


db.sqlite3
BastovOleksandr, line 1: dont push db files on git

cinema/serializers.py
```diff
@@ -1 +1,119 @@
...
+        model = Actor
+        fields = (
+            "id",
+            "first_name",
+            "last_name",
+            "full_name",
+        )
+
+    def get_full_name(self, obj):
+        return f"{obj.first_name} {obj.last_name}"
```
BastovOleksandr, line 34: you already have __str__ method in Actor model class, so you can try this in serializer:
source = "__str__"
in this  case u don't need get_full_name method in serializer
just telling, I don't think this is a problem

cinema/serializers.py
```diff
@@ -1 +1,119 @@
...
+class CinemaHallSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CinemaHall
+        fields = (
+            "id",
+            "name",
+            "rows",
+            "seats_in_row",
+            "capacity",
+        )
```
BastovOleksandr, line 46: It seems like it should be explicitly stated that this field is read only
I mean capacity = serializers.ReadOnlyField()
but I'm not sure to be honest

BastovOleksandr (approved):

BastovOleksandr (approved):

Arsen-hrynevych (approved):
ok

PR #651, opened 2024-09-29
Task Solution

achalaidiuk:


cinema/serializers.py
```diff
@@ -1 +1,75 @@
...
+from rest_framework import serializers
+
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
+
+
+class GenreSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = Genre
+        fields = "__all__"
```
DSahalatyi: If there aren't lots of fields it is better to explicitly write them as an iterable.

cinema/serializers.py
```diff
@@ -1 +1,75 @@
...
+        model = Movie
+        fields = "__all__"
+
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.SlugRelatedField(
+        many=True,
+        read_only=True,
+        slug_field="name",
+    )
```
DSahalatyi, line 38: Not a mistake. Just fyi you can use StringRelatedField here as well. So, why are you using a different field type for genres?

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+    serializer_class = ActorSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
+        if self.action == "list":
```
DSahalatyi: Are you sure you are optimizing all needed actions?

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieDetailSerializer
+        return self.serializer_class
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
```
DSahalatyi, line 46: Not needed anymore

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+
+    def get_queryset(self):
+        queryset = self.queryset
+        if self.action == "list":
```
DSahalatyi: Same as above

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+            queryset = queryset.select_related("movie", "cinema_hall")
```
DSahalatyi: Also not a mistake, but I guess there's no need to specify related fields here, especially considering you are using data from both of them.

cinema/models.py
```diff
@@ -27,8 +27,12 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
-        return self.first_name + " " + self.last_name
+        return self.full_name
```
DSahalatyi: You are setting full_name as a property and then getting it by calling str?
Only 1 of those actions should be enough.

DSahalatyi (changes requested):

DSahalatyi (approved):
I'd suggest changing all `__all__` instances.
Also, you don't need to set serializer class when you have a get_serializer_class method redefined.

cinema/serializers.py
```diff
@@ -1 +1,75 @@
-# write serializers here
+from rest_framework import serializers
+
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

Arsen-hrynevych (approved):
Overall ok

PR #653, opened 2024-09-29
solution

HalynaRohatska:


cinema/views.py
```diff
@@ -1 +1,68 @@
...
+            return MovieRetrieveSerializer
+
+        return MovieSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+
+    def get_queryset(self):
+        queryset = self.queryset
```
yashkunn: You should remove this line of code because it's redundant; the queryset is already defined at the class level, and you can directly use self.queryset in the get_queryset method.

cinema/views.py
```diff
@@ -1 +1,68 @@
...
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieListSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset
```
yashkunn: You should remove this line of code because it's redundant; the queryset is already defined at the class level, and you can directly use self.queryset in the get_queryset method.

anton-bilous (approved):

yashkunn (approved):

cinema/urls.py
```diff
@@ -1 +1,24 @@
...
+router.register("actors", ActorViewSet)
+router.register("genres", GenreViewsSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
Arsen-hrynevych, line 24: It could be simplified to urlpatterns = router.urls

Arsen-hrynevych (approved):
Thanks

PR #654, opened 2024-09-29
all done

AndrZhuk:


cinema/views.py
```diff
@@ -1 +1,65 @@
...
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieRetrieveSerializer
+        return MovieSerializer
+
+    def retrieve(self, request, *args, **kwargs):
+        instance = self.get_object()
+        serializer = self.get_serializer(instance)
+        print(type(serializer.data))
```
vsmutok: remove this `print`

cinema/urls.py
```diff
@@ -1 +1,28 @@
...
+router.register("movies", MovieViewSet, basename="movies")
+router.register(
+    "movie_sessions",
+    MovieSessionViewSet,
+    basename="movie_sessions"
+)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
vsmutok: You could simplify it to urlpatterns = router.urls

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
vsmutok: You are setting full_name as a property and then getting it by calling str?
Only 1 of those actions should be enough.

vsmutok (changes requested):

cinema/models.py
vsmutok: use  `full_name` here

cinema_service/settings.py
```diff
@@ -20,9 +20,8 @@
 # See https://docs.djangoproject.com/en/4.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = (
+SECRET_KEY = \
```
vsmutok: better to use `()` then `\`

vsmutok (changes requested):

cinema/serializers.py
```diff
@@ -1 +1,130 @@
...
+class CinemaHallSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CinemaHall
+        fields = ["name", "rows", "seats_in_row", "capacity"]
+
+
+class MovieSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = "__all__"
```
Dimosphen1, line 38: Since `__all__` is a default value, you could not include it in the future

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #659, opened 2024-09-30
implement py-api-serializers

ValeriiaKyr:


cinema/serializers.py
```diff
@@ -1 +1,85 @@
...
+from rest_framework import serializers
+
+from cinema.models import Movie, Genre, Actor, MovieSession, CinemaHall
+
+
+class ActorsSerializer(serializers.ModelSerializer):
+    full_name = serializers.SerializerMethodField()
+
+    def get_full_name(self, actor):
+        return actor.first_name + " " + actor.last_name
```
OleksiukStepan, line 10: You can simplify it by using str(actor) since the Actor model already includes this method.

cinema/serializers.py
```diff
@@ -1 +1,85 @@
...
+    def get_actors(self, movie: Movie) -> list[str]:
+        return [
+            f"{actor.first_name} {actor.last_name}"
+            for actor in movie.actors.all()
+        ]
+
+
+class MovieRetrieverSerializer(MovieSerializer):
+    genres = GenreSerializer(many=True)
+    actors = ActorsSerializer(many=True)
```
OleksiukStepan, line 52: I recommend implement this fields like read_only, because you don't need to provide genres and actors data for update movies

OleksiukStepan (approved):

cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+
+router = routers.DefaultRouter()
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("actors", ActorViewSet)
+router.register("genres", GenreViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
```
Arsen-hrynevych, line 22: [optional] It could be simplified to urlpatterns = router.urls

Arsen-hrynevych (approved):
Ok

PR #663, opened 2024-09-30
"Solution"

mnezbrytska:


HalynaRohatska (approved):

cinema/serializers.py
```diff
@@ -1 +1,89 @@
-# write serializers here
+from rest_framework import serializers
+
+from cinema.models import Movie, MovieSession, Actor, Genre, CinemaHall
```

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)

cinema/urls.py
```diff
@@ -1 +1,21 @@
...
+app_name = "cinema"
+router = routers.DefaultRouter()
+router.register("movies", MovieViewSet)
+router.register("actors", ActorViewSet)
+router.register("genres", GenreViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
Arsen-hrynevych, line 21: It could be simplified to urlpatterns = router.urls

Arsen-hrynevych (approved):
Overall ok

PR #675, opened 2024-10-02
not solution, i go to QnA tommorow coz 2 tests failed…

Ivan-Shakhman:
i realy dont undestand why so dont look at this commit before, it still not worked,

### DIFF
--- a/cinema/models.py
+++ b/cinema/models.py
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
 
--- a/cinema/models.py
+++ b/cinema/models.py
@@ -32,7 +32,7 @@ class Actor(models.Model):
         return f"{self.first_name} {self.last_name}"
 
     def __str__(self):
-        return self.first_name + " " + self.last_name
+        return self.full_name
 
 
 class Movie(models.Model):
--- a/cinema/serializers.py
+++ b/cinema/serializers.py
@@ -17,7 +17,6 @@ class GenreSerializer(serializers.ModelSerializer):
     class Meta:
         model = Genre
         fields = ("id", "name")
-        read_only_fields = ["id"]
 
 
 class ActorSerializer(serializers.ModelSerializer):
@@ -25,7 +24,6 @@ class ActorSerializer(serializers.ModelSerializer):
     class Meta:
         model = Actor
         fields = ("id", "first_name", "last_name", "full_name")
-        read_only_fields = ["id"]
 
 
 class MovieSerializer(serializers.ModelSerializer):
@@ -64,7 +62,7 @@ class CinemaHallSerializer(serializers.ModelSerializer):
     class Meta:
         model = CinemaHall
         fields = ("id", "name", "rows", "seats_in_row", "capacity")
-        read_only_fields = ["id", "capacity"]
+        read_only_fields = ["capacity"]
 
 
 class MovieSessionSerializer(serializers.ModelSerializer):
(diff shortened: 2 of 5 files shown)


cinema/serializers.py
```diff
@@ -1 +1,119 @@
...
+    Ticket
+)
+
+
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Genre
+        fields = ("id", "name")
+        read_only_fields = ["id"]
+
```
yashkunn, line 19: no need to set id field read only. this is done automatically. check other places

cinema/views.py
```diff
@@ -1 +1,54 @@
...
+
+
+class ActorViewSet(viewsets.ModelViewSet):
+    queryset = Actor.objects.all()
+    serializer_class = ActorSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+
```
yashkunn, line 24: add def get_queryset and set select_related/prefetch_related

cinema/views.py
```diff
@@ -1 +1,54 @@
...
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all().select_related(
+        "movie",
+        "cinema_hall"
+    )
+
```
yashkunn, line 49: add def get_queryset and set select_related/prefetch_related

yashkunn (approved):

cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
     def __str__(self):
         return self.first_name + " " + self.last_name
```
MNDonut: reuse fullname here

cinema/serializers.py
```diff
@@ -1 +1,116 @@
-# write serializers here
+from rest_framework import serializers
+from rest_framework.fields import SerializerMethodField, IntegerField
+from rest_framework.relations import SlugRelatedField
```
MNDonut: you don't need these imports
just use serializers.SlugRelatedField, etc..

cinema/serializers.py
```diff
@@ -1 +1,116 @@
...
+            "show_time",
+            "movie_title",
+            "cinema_hall_name",
+            "cinema_hall_capacity",
+        )
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
+    movie = MovieListSerializer(many=False)
+    cinema_hall = CinemaHallSerializer(many=False)
```
MNDonut: False is a default value, thus you can remove it

cinema/serializers.py
```diff
@@ -1 +1,116 @@
...
+    class Meta:
+        model = MovieSession
+        fields = ("id", "show_time", "movie", "cinema_hall")
+
+
+class OrderSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Order
+        fields = ("id", "created_at", "user")
+        read_only_fields = ["id", "created_at"]
```
MNDonut: id is readonly by default

cinema/views.py
```diff
@@ -1 +1,59 @@
...
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieSerializer
+
+    def get_queryset(self):
+        queryset = self.queryset.prefetch_related("genres", "actors")
+        return queryset
```
MNDonut: What is the purpose of overriding method when you can do it in queryset attribute? Also are you sure you should prefetch fields for all actions? Fix everywhere

MNDonut (changes requested):

oliasymak24 (approved):

PR #713, opened 2024-12-07
implement solution serializers folowing readme.md

Mxbely:


cinema/serializers.py
```diff
@@ -1 +1,103 @@
...
+
+class ActorSerializer(serializers.ModelSerializer):
+    full_name = serializers.SerializerMethodField(read_only=True)
+
+    class Meta:
+        model = Actor
+        fields = ["id", "first_name", "last_name", "full_name"]
+
+    def get_full_name(self, obj):
+        return f"{obj.first_name} {obj.last_name}"
```
KKvladys: U have method __str__ in model Actor and here u just repeat it

KKvladys (approved):
Gj

cinema/models.py
```diff
@@ -30,6 +30,9 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    def get_full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
```
AnyoneClown: Why didn't you use `property` here?

cinema/serializers.py
AnyoneClown: +, fix this

cinema/views.py
```diff
@@ -1 +1,85 @@
...
+
+
+class OrderSet(viewsets.ModelViewSet):
+    queryset = Order.objects.all()
+    serializer_class = OrderSerializer
+
+
+class TicketSet(viewsets.ModelViewSet):
+    queryset = Ticket.objects.all()
+    serializer_class = TicketSerializer
```
AnyoneClown: All names of Views should contain `view` word inside. Example;

`OrderViewSet`. Fix in all cases

cinema_service/urls.py
```diff
@@ -1,6 +1,8 @@
 from django.contrib import admin
-from django.urls import path
+from django.urls import path, include
+
 
 urlpatterns = [
     path("admin/", admin.site.urls),
+    path("api/cinema/", include("cinema.urls")),
```
AnyoneClown: Don't forget to add `namespace` for this URL

AnyoneClown (changes requested):

cinema/models.py
Mxbely: maybe, when i did this task, i didn't understand this moment, i will rewrite it

cinema/models.py
```diff
@@ -30,6 +30,9 @@ class Actor(models.Model):
     def __str__(self):
         return self.first_name + " " + self.last_name
 
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
+
```
AnyoneClown, line 35: This method should have decorator `@property`

AnyoneClown (approved):
Good, but check comment

PR #720, opened 2024-12-10
 Solution-py-api-serializers

skyfoxwork:


alauda6 (approved):
Don't forget about basename in router. Beside that , looks correct

PR #721, opened 2024-12-10
Solution

mshmygel:


cinema/serializers.py
```diff
@@ -1 +1,90 @@
...
+    cinema_hall_capacity = serializers.IntegerField(
+        source="cinema_hall.capacity"
+    )
+
+    class Meta:
+        model = MovieSession
+        exclude = ["movie", "cinema_hall"]
+
+
+class MovieSessionDetailSerializer(MovieSessionSerializer):
```
YaroslavBordovoy, line 76: It might be worth specifying many=False, since we have ForeignKey in our models.

YaroslavBordovoy (approved):
LGTM

PR #728, opened 2024-12-11
Solution

dxrrkwm:


molodsh1y1 (approved):
LGTM! But don't use black to all project or change your line separator

PR #772, opened 2025-02-09
Solution

taraskhom:


cinema/views.py
```diff
@@ -1 +1,65 @@
...
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieRetrieveSerializer
+        return MovieSerializer
+
+    def get_queryset(self):
+        if self.action == "list":
+            return Movie.objects.prefetch_related("genres", "actors")
+        elif self.action == "retrieve":
+            return Movie.objects.prefetch_related("genres", "actors")
```
DanilTrukhanov: u can avoid code duplication here by just writing:
        if self.action in ("list", "retrieve"):
           return Movie.objects.prefetch_related("genres", "actors")

DanilTrukhanov (approved):
Good job!

PR #773, opened 2025-02-09
Implement solution

mykytafabrykator:


mykytafabrykator:
"The capacity field is included in the CinemaHallSerializer and MovieSessionListSerializer, but it seems that the CinemaHall model does not have a capacity attribute. This will lead to errors when the serializers are used. Please ensure that the CinemaHall model includes a capacity attribute or remove it from the serializers." - capacity is an attribute of model CinemaHall, coded as a property.

cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+router = routers.DefaultRouter()
+router.register("genres", GenreViewSet)
+router.register("actors", ActorViewSet)
+router.register("cinema_halls", CinemaHallViewSet)
+router.register("movies", MovieViewSet)
+router.register("movie_sessions", MovieSessionViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
vsmutok: When basepath is empty no need to include, instead use:
urlpatterns = router.urls

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+        if self.action in ("list", "retrieve"):
+            queryset = queryset.prefetch_related("genres", "actors")
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieListSerializer
+        if self.action == "retrieve":
+            return MovieRetrieveSerializer
+        return MovieSerializer
```
vsmutok: Could you just return self.serializer_class here ?

cinema/views.py
```diff
@@ -1 +1,65 @@
...
+        if self.action in ("list", "retrieve"):
+            queryset = queryset.select_related()
+        return queryset
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+        if self.action == "retrieve":
+            return MovieSessionRetrieveSerializer
+        return MovieSessionSerializer
```
vsmutok: the same

vsmutok (changes requested):

vsmutok (approved):
Looks Good!

PR #774, opened 2025-02-09
Develop

DanilTrukhanov:


cinema/urls.py
```diff
@@ -1 +1,23 @@
...
+    GenreViewSet,
+    ActorViewSet,
+    MovieViewSet,
+    MovieSessionViewSet,
+)
+
+app_name = "cinema"
+
+router = routers.DefaultRouter()
+router.register("cinema_halls", CinemaHallViewSet)
```
4ikkamony, line 15: IDK, maybe it's worth specifying basename for views with names that consist of a few words, so that reverse url is not cinemahall but follows the convention like cinema-hall and stuff

4ikkamony (approved):
Great work!

cinema/urls.py
DanilTrukhanov: That's a good suggestion, thanks)

PR #848, opened 2025-04-16
Solution

Baranotik15:


kram3ko (approved):
Good Job.

cinema/serializers.py
```diff
@@ -1 +1,133 @@
...
+
+class ActorSerializer(serializers.ModelSerializer):
+    full_name = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Actor
+        fields = ["id", "first_name", "last_name", "full_name"]
+
+    def get_full_name(self, obj):
+        return f"{obj.first_name} {obj.last_name}"
```
Boyok1337, line 13: U can define it in models and reuse it everywhere, then it`s easy to support and change

cinema/serializers.py
```diff
@@ -1 +1,133 @@
...
+
+class ActorDetailSerializer(serializers.ModelSerializer):
+    full_name = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Actor
+        fields = ["first_name", "last_name", "full_name"]
+
+    def get_full_name(self, obj):
+        return f"{obj.first_name} {obj.last_name}"
```
Boyok1337, line 33: The same

cinema/urls.py
```diff
@@ -1 +1,22 @@
-# write urls here
+from django.urls import path, include
+from rest_framework.routers import DefaultRouter
+from .views import (
```
Boyok1337, line 3: Absolute imports better then relative

cinema_service/settings.py
```diff
@@ -20,9 +20,8 @@
 # See https://docs.djangoproject.com/en/4.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = (
+SECRET_KEY = \
```
Boyok1337, line 23: What is that?

Boyok1337 (changes requested):
If u use black, format only files that you created, but not a whole project

PR #854, opened 2025-04-21
finish

PrimeGlorious:


cinema/views.py
```diff
@@ -1 +1,88 @@
...
+                "genres",
+                "actors"
+            )
+        elif self.action == "retrieve":
+            return Movie.objects.all().prefetch_related(
+                "genres",
+                "actors"
+            )
+
+        return Movie.objects.all()
```
Boyok1337, line 59: If u overwrite get_queryset, u don`t need to set `queryset = Movie.objects.all()` 
And u can just use prefetch_related with no if elif construction

Boyok1337 (approved):
Approve, but check and fix comments

PR #901, opened 2025-06-06
Develop

MAksum123456:


cinema/urls.py
```diff
@@ -1 +1,31 @@
...
+router.register("movie_sessions", MovieSessionViewSet)
+router.register("orders", OrderViewSet)
+router.register("tickets", TicketViewSet)
+
+
+app_name = "cinema"
+
+urlpatterns = [
+    path("", include(router.urls))
+]
```
vsmutok: When basepath is empty no need to include, instead use:
urlpatterns = router.urls

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #950, opened 2025-08-31
'Deal'

DanilHamster:


cinema/models.py
```diff
@@ -27,6 +27,10 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    @property
+    def full_name(self):
+        return f"{self.first_name} {self.last_name}"
```
AnyoneClown, line 32: No annotation for `return`

AnyoneClown (approved):
LGTM to me, but don't forget about type hints

PR #965, opened 2025-09-03
solution

Sashoxxx:


cinema/serializers.py
```diff
@@ -1 +1,89 @@
...
+        fields = ("id", "name", "rows", "seats_in_row", "capacity")
+
+
+class MovieSerializer(serializers.ModelSerializer):
+    genres = serializers.PrimaryKeyRelatedField(
+        many=True, queryset=Genre.objects.all()
+    )
+    actors = serializers.PrimaryKeyRelatedField(
+        many=True, queryset=Actor.objects.all()
+    )
```
yagginov, line 35: You don't need to write this for primary_key, it will pull it up on its own, the main thing is to specify fields.

cinema/serializers.py
```diff
@@ -1 +1,89 @@
...
+
+class ActorSerializer(serializers.ModelSerializer):
+    full_name = serializers.SerializerMethodField(read_only=True)
+
+    class Meta:
+        model = Actor
+        fields = ("id", "first_name", "last_name", "full_name")
+
+    def get_full_name(self, actor: Actor) -> str:
+        return f"{actor.first_name} {actor.last_name}"
```
yagginov, line 20: I think it would have been better to use properties on the model.

yagginov (approved):
well

Sashoxxx:
@mateacademy-ai-mentor edited queryset, and added annotation, don't need to use **SerializerMethodField ** for full name, as Actor model provides us proper __str__ method.

Sashoxxx:
@mateacademy-ai-mentor   Actor model defines str method. look here [https://github.com/mate-academy/py-api-serializers/blob/master/cinema/models.py](url). All test are passed.

PR #966, opened 2025-09-03
Solution

yagginov:


Kostenikov (approved):
┏ ( ゜ω゜)=☞YOU CAN DO BETTER❗❗❗

yagginov:
@mateacademy-ai-mentor 
1. Movie list actors as strings
This is not an issue. Since the field is read_only=True, DRF simply calls getattr(actor, "full_name") and the property works fine. The current implementation exactly matches the required output format (actors as strings), so changing to SerializerMethodField adds boilerplate without any benefit.

2. MovieSession detail must nest MovieDetailSerializer
This is incorrect. The README explicitly requires genres and actors as strings, not as nested objects. Using MovieListSerializer is the correct choice here because it outputs exactly the structure specified in the task. Switching to MovieDetailSerializer would break the required format.

yagginov:
@mateacademy-ai-mentor 

1. Movie list actors as strings
This is not an issue. Since the field is read_only=True, DRF simply calls getattr(actor, "full_name") and the property works fine. I USE READ_ONLY so it`s works right without any problems.

2. MovieSession detail must nest MovieDetailSerializer
This is incorrect. The README explicitly requires genres and actors as strings:
```
HTTP 200 OK
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "id": 1,
    "show_time": "2022-06-15T12:25:00Z",
    "movie": {
        "id": 1,
        "title": "Harry Potter and the Philosopher's Stone",
        "description": "The first movie about Harry Potter",
        "duration": 210,
        "genres": [
            "drama"
        ],
        "actors": [
            "Emma Watson",
            "Daniel Radcliffe"
        ]
    },
    "cinema_hall": {
        "id": 2,
        "name": "Black",
        "rows": 20,
        "seats_in_row": 15,
        "capacity": 300
    }
}
```
(this i take from readme file), not as nested objects. Using MovieListSerializer is the correct choice here because it outputs exactly the structure specified in the task. Switching to MovieDetailSerializer would break the required format.

yagginov:
@mateacademy-ai-mentor 

I took this from the pull request you are referring to, it's one-to-one with my code. ([#3](https://github.com/mate-academy/py-api-serializers/pull/3)):
```
class MovieListSerializer(MovieSerializer):
    genres = serializers.SlugRelatedField(
        many=True,
        read_only=True,
        slug_field="name"
    )
    actors = serializers.SlugRelatedField(
        many=True,
        read_only=True,
        slug_field="full_name"
    )
```
    
What will you say now?

PR #967, opened 2025-09-03
Solution

Kostenikov:


cinema/serializers.py
```diff
@@ -1 +1,84 @@
...
+
+class MovieSerializer(serializers.ModelSerializer):
+    actors = serializers.PrimaryKeyRelatedField(
+        many=True,
+        queryset=Actor.objects.all()
+    )
+    genres = serializers.PrimaryKeyRelatedField(
+        many=True,
+        queryset=Genre.objects.all()
+    )
```
dimashevchukk, line 32: not needed i believe

cinema/views.py
```diff
@@ -1 +1,51 @@
...
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+
+    def get_serializer_class(self):
+        if self.action == "list":
+            return MovieSessionListSerializer
+        if self.action == "retrieve":
+            return MovieSessionDetailSerializer
+        return MovieSessionSerializer
```
dimashevchukk: could be N+1, add get_queryset

dimashevchukk (approved):

Kostenikov:
@mateacademy-ai-mentor you said:
> This violates the checklist and requirements: SlugRelatedField with slug_field="full_name" will not work because full_name is a property, not a model field. You should use a SerializerMethodField to return the actor's full name as required by the Movie list endpoint output.

You are wrong, we don't have this in checklist and i don't use full_name in post-put-patch operations so property will work correct with SlugRelatedField and read_only=True, nothing to change here.

Other issues was fixed.

PR #978, opened 2025-09-03
Solution

Arch0998:


Salmon-eat (approved):
good

Arch0998:
@mateacademy-ai-mentor According to the README and the provided tests, my implementation is correct and works as expected. The serializers and API responses match the requirements and test cases, so any claim of an error is mistaken. All tests pass and the project behaves according to the documentation.

Arch0998:
@mateacademy-ai-mentor 
You write: How to fix: Change the field to use the detail serializer.
Replace: movie = MovieSerializer(read_only=True)
With: movie = MovieDetailSerializer(read_only=True)

But if I do that, the tests will fail. This won’t solve the problem.

PR #979, opened 2025-09-04
solution

trellle:


MakovM (changes requested):
There is no solution, you must complete the task.

PR #981, opened 2025-09-04
compled task api serializers

Salmon-eat:


cinema/views.py
```diff
@@ -1 +1,59 @@
...
+class CinemaHallViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MovieViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+    def get_serializer_class(self):
```
trellle: implement please get_queryset methods in order to optimize work with database

cinema/views.py
```diff
@@ -1 +1,59 @@
...
+        if self.action == "list":
+            return MovieSerializer
+        if self.action == "retrieve":
+            return MovieDetailSerializer
+        return MovieWriteSerializer
+
+
+class MovieSessionViewSet(viewsets.ModelViewSet):
+    queryset = MovieSession.objects.all()
+    serializer_class = MovieSessionSerializer
```
trellle, line 58: get_queryset method also would be useful here

cinema/serializers.py
```diff
@@ -1 +1,124 @@
...
+class ActorSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Actor
+        fields = ("id", "first_name", "last_name", "full_name")
+
+
+class GenreSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Genre
+        fields = ("id", "name")
```
trellle, line 20: id field should be read only

trellle (changes requested):
please improve yo solution

cinema/serializers.py
Salmon-eat: I don't see the point.

PR #984, opened 2025-09-07
Solution

Sir-Churchill:


Sir-Churchill:
2. In model we define by property this filed so i don`t see any sense about create new logic if i have it in model
3.Exactly same situation i have __str__ in model and i can call full_name by StringRelatedField so why i need write more code?
4.How i see example and my result it exactly the same and it creates all how in README 
5.What point of making update and create in ModelSerializer if it this provided in serializer and it automatically works with update and create so 0 logic for this
6. Queryset can`t be NONE if we delete this from class it destroy logic by drf

Sir-Churchill:
2.I have it in models like property what calculate what i need why i need overwrite and make more if all works correctly
3.For what reason and need to write more code if i have it models and use it
4.My Selializer works correct like in readme and makes full_name for actors and cinema_hall exactly the same like in example
5.For what reason i need to rewrite update and create if i have it in ModelViewSet and it works correct
6.queryset can`t be NONE in class because i only add functionality extra for my situation and if i delete it, it will destroy logic

PR #1038, opened 2025-12-08
Solution

PythonBossUA:


cinema/urls.py
```diff
@@ -1 +1,27 @@
...
+
+router.register("cinema_halls", CinemaHallsViewSet)
+
+router.register("movies", MoviesViewSet)
+
+router.register("movie_sessions", MovieSessionsViewSet)
+
+urlpatterns = [
+    path("", include(router.urls)),
+]
```
vsmutok:
```suggestion
urlpatterns = router.urls
```

cinema/views.py
```diff
@@ -1 +1,62 @@
...
+
+class CinemaHallsViewSet(viewsets.ModelViewSet):
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class MoviesViewSet(viewsets.ModelViewSet):
+    queryset = Movie.objects.all()
+
+    def get_queryset(self):
```
vsmutok: def get_queryset(self): is missing a type annotation on the method signature. The checklist requires type annotations in class methods. Add an appropriate return type (for example -> QuerySet). Quote: "Do not forget about type annotation in class methods."

.flake8
```diff
@@ -1,7 +1,7 @@
 [flake8]
 inline-quotes = "
 ignore = E203, E266, W503, N807, N818, F401
-max-line-length = 79
+max-line-length = 100
```
vsmutok: Please do not change this.
Revert it back to its original state.

vsmutok (changes requested):

DaTrEvTeR (approved):
Well done!

PR #1051, opened 2025-12-28
Solution

stanislav-kishechko:


cinema/serializers.py
```diff
@@ -1 +1,94 @@
-# write serializers here
+from rest_framework import serializers
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
```
Vorongor: As you said one class - one row)

cinema/views.py
```diff
@@ -1 +1,53 @@
-# write views here
+from rest_framework import viewsets
+
+from cinema.models import Genre, Actor, CinemaHall, Movie, MovieSession
```
Vorongor: same)

Vorongor (approved):
Well Done!

PR #1055, opened 2025-12-29
Soullution

Iventyk:


vakt159 (approved):
Don't know about  predefining querysets with already select_related() and prefetch_related() in Views. Everything else looks good

PR #1061, opened 2025-12-29
Develop

AngryAndrii:


cinema/serializers.py
```diff
@@ -1 +1,149 @@
-# write serializers here
+from rest_framework import serializers
+
+from cinema.models import CinemaHall, Genre, Actor, Movie, MovieSession
```
tigersaymon: Use parentheses here for better readability

cinema/serializers.py
```diff
@@ -1 +1,149 @@
...
+        return f"{obj.first_name} {obj.last_name}"
+
+
+class MovieSerializer(serializers.ModelSerializer):
+    genres = serializers.SlugRelatedField(
+        many=True,
+        read_only=True,
+        slug_field="name"
+    )
+    actors = serializers.SerializerMethodField()
```
tigersaymon, line 41: I think we can achieve the same result with fewer lines if we do 
actors = serializers.StringRelatedField(many=True, read_only=True)

cinema/serializers.py
```diff
@@ -1 +1,149 @@
...
+        fields = ("id", "title", "description", "duration", "genres", "actors")
+
+
+class MovieListSerializer(MovieSerializer):
+    genres = serializers.SlugRelatedField(
+        many=True,
+        read_only=True,
+        slug_field="name"
+    )
+    actors = serializers.SerializerMethodField()
```
tigersaymon, line 80: DRY
You inherit MovieListSerializer from MovieSerializer. In the parent MovieSerializer, you have already written the logic for genres (SlugRelated) and actors (MethodField). In the child MovieListSerializer, you completely copy this code. This is unnecessary.

cinema/serializers.py
```diff
@@ -1 +1,149 @@
...
+
+    class Meta:
+        model = MovieSession
+        fields = (
+            "id",
+            "show_time",
+            "movie_title",
+            "cinema_hall_name",
+            "cinema_hall_capacity",
+        )
```
tigersaymon, line 121: Inheriting from MovieSessionSerializer causes an issue here. The parent serializer already includes the full nested movie and cinema_hall objects. By adding movie_title and cinema_hall_name on top, the API response will contain duplicate data (both the full objects and the flat fields).

cinema/views.py
```diff
@@ -1 +1,53 @@
...
+    queryset = CinemaHall.objects.all()
+    serializer_class = CinemaHallSerializer
+
+
+class GenreViewSet(viewsets.ModelViewSet):
+    queryset = Genre.objects.all()
+    serializer_class = GenreSerializer
+
+
+class ActorsViewSet(viewsets.ModelViewSet):
```
tigersaymon: Please rename this to ActorViewSet (singular)

cinema/views.py
```diff
@@ -1 +1,53 @@
-# write views here
+from typing import Type
+
+from rest_framework import viewsets
+from rest_framework.serializers import BaseSerializer
+
+from cinema.models import Genre, CinemaHall, Actor, Movie, MovieSession
```
tigersaymon: Could you also run a formatter on the imports?

tigersaymon (changes requested):

tigersaymon (approved):

PR #1194, opened 2026-06-16
first solution but two tests are failing

heyimfles:


cinema/views.py
```diff
@@ -1 +1,102 @@
...
+            )
+
+        return queryset
+
+    def get_serializer_class(self) -> type[Serializer]:
+        if self.action == "list":
+            return MovieSessionListSerializer
+        elif self.action == "retrieve":
+            return MovieSessionRetrieveSerializer
+        else:
```
vsmutok: `else` redundant in this case

cinema/views.py
```diff
@@ -1 +1,102 @@
...
+            return MovieListSerializer
+        elif self.action == "retrieve":
+            return MovieRetrieveSerializer
+        elif (
+            self.action == "create"
+            or self.action == "update"
+            or self.action == "partial_update"
+        ):
+            return MovieCreateUpdateSerializer
+        else:
```
vsmutok: `else` redundant in this case

cinema/views.py
```diff
@@ -1 +1,102 @@
...
+from cinema.serializers import (
+    GenreSerializer,
+    ActorSerializer,
+    CinemaHallSerializer,
+    MovieSerializer,
+    MovieSessionSerializer,
+    MovieListSerializer,
+    MovieSessionListSerializer,
+    MovieSessionRetrieveSerializer,
+    MovieRetrieveSerializer, MovieCreateUpdateSerializer,
```
vsmutok:
```suggestion
    MovieSessionRetrieveSerializer,
    MovieRetrieveSerializer, 
    MovieCreateUpdateSerializer,
```

vsmutok (changes requested):

cinema/views.py
heyimfles: sorry, i don't really undertsand what needs the change here. should i do this

```
        if self.action == "list":
            return MovieSessionListSerializer

        if self.action == "retrieve":
            return MovieSessionRetrieveSerializer

        return MovieSessionSerializer
```

or just remove the regular MovieSessionSerializer ?

.gitignore
```diff
@@ -6,3 +6,6 @@
 venv/
 .pytest_cache/
 **__pycache__/
+db.sqlite3
+3.15
+5.1
```
vsmutok: why you add `3.15`, `5.1` to `.gitignore` file?

cinema/urls.py
```diff
@@ -1 +1,47 @@
...
+router = routers.DefaultRouter()
+
+router.register(
+    "genres",
+    GenreViewSet
+)
+router.register(
+    "actors",
+    ActorViewSet
+)
```
vsmutok:
```suggestion
router.register("genres", GenreViewSet)
router.register("actors", ActorViewSet)
```

**If you can fit this into a single line, it's better to do so. That way, the code will be easier to read.**

vsmutok (changes requested):

3.15
vsmutok, line 1: remove this file from the PR

5.1
vsmutok, line 1: remove this file from the PR

vsmutok (changes requested):

vsmutok (approved):
Great!

